PR #252, opened 2014-12-27
Shorten Modifier syntax with `with`

reem:
Modifiers are more flexible than the custom constructors we used to have, but came at the cost of more verbose initialization, see:

Before: `Response::with(status::Ok, "Hello World!")`
After: `Response::new().set(status::Ok).set("Hello World!")`

But by implementing `Modifier<X>` for tuples of `Modifier<X>` and a new `Response::with` constructor which pre-applies a modifier we can recover this brevity:

Now: `Response::with((status::Ok, "Hello World!"))`

Though slightly longer due to the tuple construction, this is a huge improvement over the original `with` since it is _far_ more flexible, for instance:
- `Response::with(status::NotFound)`
- `Response::with((status::Ok, Path::new("some_file")))`
- `Response::with((status::Ok, json::encode(data), json_mime_type))`

all work out of the box.

### DIFF
--- a/examples/404.rs
+++ b/examples/404.rs
@@ -6,7 +6,7 @@ use iron::status;
 
 fn main() {
     Iron::new(|&: _: &mut Request| {
-        Ok(Response::new().set(status::NotFound))
+        Ok(Response::with(status::NotFound))
     }).listen("localhost:3000").unwrap();
     println!("On 3000");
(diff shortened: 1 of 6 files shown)


barosl:
:+1: :clap: :star: Nice work!

PR #254, opened 2015-01-02
(fix) Update to use new array syntax.

Nercury:
[New array syntax](https://github.com/rust-lang/rust/pull/20387).


reem:
Thanks.

PR #256, opened 2015-01-11
Update to rust nightly 2015-01-09

pzol:
This fixes all but those errors:

src/response/modifiers.rs:37:1: 42:2 error: the type `mime::Mime` does not reference any types defined in this crate; only traits defined in the current crate can be implemented for arbitrary types [E0117]
src/response/modifiers.rs:37 impl Modifier<Response> for Mime {
src/response/modifiers.rs:38     #[inline]
src/response/modifiers.rs:39     fn modify(self, res: &mut Response) {
src/response/modifiers.rs:40         res.headers.set(headers::ContentType(self))
src/response/modifiers.rs:41     }
src/response/modifiers.rs:42 }
src/response/modifiers.rs:47:1: 52:2 error: type parameter `B` is not constrained by any local type; only traits defined in the current crate can be implemented for a type parameter
src/response/modifiers.rs:47 impl<B: Bodyable> Modifier<Response> for B {
src/response/modifiers.rs:48     #[inline]
src/response/modifiers.rs:49     fn modify(self, res: &mut Response) {
src/response/modifiers.rs:50         self.set_body(res);
src/response/modifiers.rs:51     }
src/response/modifiers.rs:52 }
src/response/modifiers.rs:47:1: 52:2 note: for a limited time, you can add `#![feature(old_orphan_check)]` to your crate to disable this rule
src/response/modifiers.rs:47 impl<B: Bodyable> Modifier<Response> for B {
src/response/modifiers.rs:48     #[inline]
src/response/modifiers.rs:49     fn modify(self, res: &mut Response) {
src/response/modifiers.rs:50         self.set_body(res);
src/response/modifiers.rs:51     }
src/response/modifiers.rs:52 }
src/response/modifiers.rs:122:1: 126:2 error: the type `hyper::status::StatusCode` does not reference any types defined in this crate; only traits defined in the current crate can be implemented for arbitrary types [E0117]
src/response/modifiers.rs:122 impl Modifier<Response> for status::Status {
src/response/modifiers.rs:123     fn modify(self, res: &mut Response) {
src/response/modifiers.rs:124         res.status = Some(self);
src/response/modifiers.rs:125     }
src/response/modifiers.rs:126 }
error: aborting due to 3 previous errors
Could not compile `iron`.

### DIFF
--- a/examples/around.rs
+++ b/examples/around.rs
@@ -43,11 +43,11 @@ impl<H: Handler> Handler for LoggerHandler<H> {
 
 impl AroundMiddleware for Logger {
     fn around(self, handler: Box<Handler + Send + Sync>) -> Box<Handler + Send + Sync> {
-        box LoggerHandler {
+        Box::new(LoggerHandler {
             logger: self,
             handler: handler
         } as Box<Handler + Send + Sync>
-    }
+    })
 }
 
 fn hello_world(_: &mut Request) -> IronResult<Response> {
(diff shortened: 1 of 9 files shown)


reem:
@pzol Thanks for this, but I ended up needing to make a bunch of other unrelated changes, so I just rolled everything into #257 instead.

PR #257, opened 2015-01-12
Fix for latest Rust

reem:
This fixes everything except a series of ICEs when compiling
all example. Small victories.


reem:
Examples are ICEing, nothing that can be done right now so merging.

Issue #263, opened 2015-01-20
Files upload


s-panferov:
Hello @reem, have you already thought about support of file uploads with `multipart/form-data`? Maybe you have some ideas how it can be implemented the best way.

It looks like @cybergeek94 already made [multipart](https://github.com/cybergeek94/multipart) library for Hyper, but he has some [issues with Iron integration](https://github.com/cybergeek94/multipart/issues/1). What do you think about it?

cybergeek94:
I haven't been maintaining multipart lately due to lack of interest but I definitely wouldn't mind it becoming a component of Iron.

The main issue is that Iron provides the request body to its middleware as `Vec<u8>` which is kind of a bad idea if you want to support large request bodies, such as with file uploads. I shared some ideas in a Reddit discussion a while back but I don't know if anything came of that.

reem:
It's possible that `Request` could be changed to allow streaming of the input data. I've thought about it and the issues that come up with direct `Response` streaming aren't really an issue for `Request`, since there is only one state (i.e. no "status line written", "headers written", "nothing written" etc.). 

The major issue I anticipate is that this makes testing a bit trickier, but from my memory of the internals of Hyper I think it's actually pretty easy to mock an `HttpWriter` with whatever data you want. This also has the side-effect of making `Request` no longer `'static` or `Send` (since `HttpWriter` is given to us by Hyper with a lifetime), which can be a pain.

reem:
I can take a crack at making this change tomorrow, but it's not too complex a change and probably interesting if either of you wants to give it a try first.

reem:
We now allow streaming the request body, so Iron should now be able to support this use case well.

PR #266, opened 2015-01-21
(feat) Stream the body of a Request as a Reader

### LINKED ISSUE #265: Allow streaming of the request body
cc #263

This is really a necessary improvement to support common operations like file upload and many other use cases.

reem:
This changes the behavior of Iron in a rather crucial way.
Instead of reading the entire body of an incoming Request into a
`Vec<u8>` at the creation time of a Request, the body is now
exposed as a Reader on the Request object and can be read on-demand
or not-at-all depending on its use.

Streaming is critical for uses such as file uploading, where loading
the entire request body into memory unconditionally is... bad.

The body is exposed as a mostly opaque `Body` type which implements
`Reader` and wraps hyper's `HttpReader<&'a mut (Reader + 'a)>`. The http
reader cannot currently be retrieved from a `Body`, but for testing
purposes `Body::new` can be used to construct a `Body` from an `HttpReader`.

Progress on #263
Fixes #265

### DIFF
--- a/src/middleware/test.rs
+++ b/src/middleware/test.rs
@@ -1,11 +1,15 @@
 use std::io::net::ip::ToSocketAddr;
+use std::io::util::NullReader;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering::Relaxed;
 use std::sync::Arc;
 
+use hyper::http::HttpReader;
+
 use self::Kind::{Fine, Prob};
 
 use prelude::*;
+use request::Body;
 use {method, headers};
 use {AfterMiddleware, BeforeMiddleware, Handler, TypeMap, Url};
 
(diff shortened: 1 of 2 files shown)


reem:
cc @s-panferov @cybergeek94

I had some free time so I just made the change. Thoughts on the approach? Do you think `Request` should implement `Reader` itself, or is the current isolation to the `body` field appropriate?

cybergeek94:
I'm concerned that a naively written `BeforeMiddleware` would be able to opaquely invalidate a request by reading its body to EOF, since it is passed `&mut Request`, thus making the body unavailable to subsequent middleware and handlers. It's mostly the user's problem, but it also seems like something we could easily prevent in the API. 

Is there any use-case where `BeforeMiddleware` has to inspect the request body directly? That would break anyways with this change, since the ordering of the middleware affects how much of the body might have been read.

Perhaps `BeforeMiddleware` can get a `RequestInfo` struct that provides everything but the body, or separate the body from `Request` and pass it to `Handler` directly.

s-panferov:
I think everything is good. If a middleware wants to read the entire body, it can provide the body or it parts as request extensions (with help of TypeMap).

reem:
@cybergeek94 I think it would actually be the role of a Plugin to properly read the request body and transform it into whatever form a Handler can actually use and storing it the extensions, whether that's reading the whole thing and parsing it into JSON or parsing it using multipart or whatnot. It should be rare for the Handler to be responsible for dealing with a raw request body directly.

BeforeMiddleware used to fill this role, but the lazily-evaluated and cached nature of Plugins actually makes them ideal for dealing with this sort of one-shot data.

reem:
Updated for #275, landing.

PR #270, opened 2015-01-22
(fix) Fixed bug in redirect code

ProtectedMode:
This makes redirects work again.

### DIFF
--- a/src/modifiers.rs
+++ b/src/modifiers.rs
@@ -113,7 +113,7 @@ pub struct Redirect(pub Url);
 impl Modifier<Response> for Redirect {
     fn modify(self, res: &mut Response) {
         let Redirect(url) = self;
-        res.headers.set(headers::Location(format!("{:?}", url)));
+        res.headers.set(headers::Location(url.to_string()));
     }
 }
 


src/modifiers.rs
```diff
@@ -113,7 +113,7 @@ pub struct Redirect(pub Url);
 impl Modifier<Response> for Redirect {
     fn modify(self, res: &mut Response) {
         let Redirect(url) = self;
-        res.headers.set(headers::Location(format!("{:?}", url)));
+        res.headers.set(headers::Location(format!("{}", url)));
```
s-panferov: Maybe we can use just `to_string()` here? Or it is a bad practice?

ProtectedMode: @s-panferov, Url doesn't have a to_string method as far as I know. If it had/has, to_string would be better.

s-panferov: @ProtectedMode, it must be there, because this impl in http://doc.rust-lang.org/nightly/std/string/trait.ToString.html

```rust
impl<T> ToString for T where T: String, T: ?Sized
```

ProtectedMode:
@s-panferov, you were right about to_string, I didn't know it was implemented for these types; I changed it now.

reem:
@ProtectedMode Awesome.

Issue #272, opened 2015-01-23
Simplify the error system and associated middleware constructs


reem:
The current error system is extremely complex, and I've found myself very dissatisfied with the situation. Things are much too difficult to understand and the implementation is under-specced and in all likelihood buggy.

The new system will mostly ditch the idea of "handling" errors in favor of just returning an alternate Response and getting on to other non-broken requests. More details to come either in this issue or in an implementation.

PR #273, opened 2015-01-23
(fix) Include port in URLs again.

### LINKED ISSUE #245: Error handling shows wrong port
I've got a server running on localhost:3000, yet going to a URL that doesn't have a route defined shows:

```
ERROR:iron::iron: Error handling:
Request {
    url: http://localhost:80/goodbye
    method: GET
    remote_addr: 127.0.0.1:48341
    body: []
}
Error was: (HTTP/1.1 500 Internal Server Error Internal Server Error

, Err(NoRoute))
```

Note port 80 in the URL.

As a guess, perhaps something isn't passing the port along, so `from_generic_url` is defaulting to 80.

michaelsproul:
Closes #245.

Not sure if this is the best approach. The alternative is modifying Hyper so that it includes the local `SocketAddr` in its request struct. You'd probably be a better judge of this @reem, given your involvement with Hyper.

### DIFF
--- a/src/iron.rs
+++ b/src/iron.rs
@@ -2,7 +2,6 @@
 //! `Iron` library.
 
 use std::io::net::ip::{ToSocketAddr, SocketAddr};
-use std::io::{Listener};
 
 pub use hyper::server::Listening;
 use hyper::server::Server;
@@ -22,6 +21,9 @@ pub struct Iron<H> {
     /// Iron contains a `Handler`, which it uses to create responses for client
     /// requests.
     pub handler: H,
+
+    /// Once listening, the local address that this server is bound to.
+    pub addr: Option<SocketAddr>
 }
 
 impl<H: Handler> Iron<H> {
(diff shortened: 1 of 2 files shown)


src/iron.rs
```diff
@@ -33,14 +35,14 @@ impl<H: Handler> Iron<H> {
     ///
     /// Defaults to a threadpool of size 100.
     pub fn listen<A: ToSocketAddr>(self, addr: A) -> IronResult<Listening> {
-        let SocketAddr { ip, port } = try!(addr.to_socket_addr());
-
-        Ok(try!(Server::http(ip, port).listen(self)))
+        self.listen_with(addr, 100)
```
michaelsproul: Just did this to follow the docs and avoid duplication.

reem: Yup. I have a similar fix in a local branch addressing #272

reem:
`hyper::server::Request` has a `remote_add` field which contains the `SocketAddr` of the request

reem:
by the way thanks for reaching out on #245, I let that one slip through the cracks

michaelsproul:
The `remote_addr ` is the IP and port of the client, is it not? From poking around in Hyper it seems the local equivalent is ignored in the `listen` methods.

reem:
You're right, that's the client address. If we're going to do this, we might as well just add the host socket address to the Request type and use that, rather than passing it around separately.

Thanks,
         Jonathan Reem

michaelsproul:
Just to clarify, you mean Hyper's request type?

Otherwise the only way we can store it on our Request type is by storing it on `Iron` as I've done here. This is because the only arguments to `hyper::server:Handler::handle` are `self` and a Hyper request + response.

reem:
Ya, just store it on Iron and add it to our Request type, we can upstream
later.

michaelsproul:
Ok, sounds sensible. Keeps our API consistent at least.

michaelsproul:
Done, but Hyper and a few other things are broken because of the change to the `std::error::Error` trait (the detail now has to be provided via `Display`).

reem:
Looks good - I'll merge when everything builds again.

michaelsproul:
Bump. It builds!

reem:
Excellent!

PR #274, opened 2015-01-24
(fix) Update to the latest rustc (4874ca36f 2015-01-23)

SkylerLipthay:
* Replacing `fmt::Show` with `fmt::Debug`, `fmt::String` with `fmt::Display`.
* Replacing `str::StrExt::slice_from` with `&s[start..]` notation.


reem:
Awesome! Thanks for doing this.

Issue #239, opened 2014-11-26
Allow several handlers


s-panferov:
Hello @reem. It's cool that you moved Iron to Hyper. So I am happy to say that I am moving [rustless](https://github.com/rustless/rustless) to Iron.

During the process of porting I found several things that disappoint me a little bit. The first one is support of only one `Handler` for application. I came from Ruby world and I saw a lot of projects that use both Rails and Sinatra/Grape. Rails serves regular HTTP-requests for HTML pages and Grape/Sinatra provides an API.

So I think that we need to extend the current implementation of `Chain`. Maybe you would want to look at [this](https://github.com/rustless/rustless/blob/master/src/middleware.rs) implementation. There is a special error (`NotMatchError `) to say that the handler didn't match and we can go to the next.

Maybe we can just add an another instance that will iterate through `Handlers` and implement `Handler` itself. What do you think about it?

reem:
Closing, unless you have a further action item.

PR #275, opened 2015-01-25
Completely rewrite the Error system

### LINKED ISSUE #186: Weirdness with ChainBuilder.catch
An error returned by ChainBuilder.call causes all the middleware to be run twice... Not ideal. Perhaps we should just remove ChainBuilder.catch?

### LINKED ISSUE #272: Simplify the error system and associated middleware constructs

reem:
Lots and lots of changes here - the commits are pretty linear
and should have good descriptions, you can also just read the
docs for IronError and the middleware module, which explain the
new situation pretty well.

Also big changes to the README, which finally got a much needed
cleanup.

Chain the trait is gone, ChainBuilder => Chain

[breaking-change]

Fixes #272
Fixes #186


reem:
Yay breaking all the things ever!

reem:
@michaelsproul @zzmp thoughts?

src/middleware.rs
```diff
@@ -16,76 +16,46 @@
...
+//! by the Handler, which can be an arbitrary nesting of AroundMiddleware, then all
+//! AfterMiddleware are called with both the Request and Response. After all AfterMiddleware
+//! have been fired, the response is written back to the client.
 //!
-//! Not all too surprising. Of note is that AfterMiddleware and BeforeMiddleware
-//! are completely separate, unlike in many other web frameworks.
+//! Iron's error handling system is pragmatic and focuses on tracking two pieces
+//! of information for error receivers (other middleware):
+//!   - The cause of the error
+//!   - The result (what to do about) the error.
```
michaelsproul: This list look funky in the HTML docs. Use a `*`?

src/middleware.rs
```diff
@@ -94,24 +64,16 @@ pub trait Handler: Send + Sync {
...
+    fn before(&self, _: &mut Request) -> IronResult<()> { Ok(()) }
 
-    /// Try to `catch` an error thrown by this Middleware or a previous `BeforeMiddleware`.
-    ///
-    /// Should only return Ok(()) if the error has been completely handled and a Chain
-    /// can proceed as normal.
-    fn catch(&self, _: &mut Request, err: IronError) -> IronResult<()> {
-        Err(err)
-    }
+    /// Try to catch an error thrown by a previous `BeforeMiddleware`.
```
reem: Should be changed to reflect the comment on AfterMiddleware::catch

reem:
Particular open questions: is it correct for `Handler` to not have a `catch`, and therefore be bypassed in the error flow?

src/middleware.rs
```diff
@@ -16,76 +16,46 @@
...
+//! When an error is thrown in Iron by any middleware or handler returning an `Err`
+//! variant with an `IronError`, the flow of the Request switches to the error flow,
+//! which proceeds to just call the `catch` method of middleware and sidesteps the
+//! `Handler` entirely, since there is already a Response in the error.
 //!
-//!   * Errors should persist and be propagated along the same route as successes until they are
-//!     handled.
+//! Once a Request has entered the error flow, it is impossible to exit. It is therefore
+//! impossible to fully "handle" an error and return to the normal flow, since there is
+//! often no sensical place to return to and intermingling the two flows results in
```
michaelsproul: Nitpick: "sensible" instead of "sensical"? I love nonsensical as a word, but sensical doesn't seem to be a word, or at least not a recognised/common word.

michaelsproul:
The system looks good, I think not attempting to handle errors is much cleaner.

I also think not having a `catch` method on AroundMiddleware is fine, as AroundMiddleware can already modify errors that occur in their wrapped handler(s) and it doesn't make much sense to go through the primary response construction pipeline if something goes badly wrong. I imagine that the catch method would be useful for communicating between related middleware, did you have any concrete examples in mind?

src/iron.rs
```diff
@@ -52,49 +63,41 @@ impl<H: Handler> Iron<H> {
     pub fn new(handler: H) -> Iron<H> {
         Iron { handler: handler }
     }
+
+    fn bad_request(&self, mut http_res: HttpResponse<Fresh>) {
```
s-panferov: Do we need to specify `Fresh`? It's default as I can see.

src/lib.rs
```diff
@@ -33,8 +33,9 @@ extern crate hyper;
 extern crate "typemap" as tmap;
 extern crate plugin;
 extern crate "modifier" as modfier;
-extern crate error;
+extern crate "error" as err;
```
s-panferov: Maybe it's better to rename `error` to something else? I usually rename it too in my own code  [same thing](https://github.com/rustless/valico/commit/e40f668b893aee50e4ba18759672c373f380e735?diff=unified#diff-dd5f1b0c97a16f836553c88faac50f8aR3).

s-panferov:
The changes look good for me. I think that this PR is a huge improvement. :+1: 

What do you think about simplify it even more? Current flow is still complicated, because:

1. We need to think about indexes of middlewares with `catch` methods.
2. Catch signature is the same in all the middleware types, and I don't know which type I need to select to handle some common error which can happen everywhere.

I propose to remove `catch` method from middleware and introduce new middleware type: `CatchMiddleware` (maybe `middleware` is the bad name for that thing, maybe just `Catcher` is better) with one `catch` method. When error fires we just pass it through *all the catchers*. It's much simpler for me to understand this flow with `Catchers`: Need to handle some specific error? Just place the Catcher to the catcher list.

What do you think? Do you see some disadvantages of that flow?

src/lib.rs
reem: Hmm.. I think it's usually not needed - I only did it here so I could have a module named error as well.

reem:
@s-panferov There are a few disadvantages to splitting everything up:

1. You have to link stuff like loggers an additional time
2. Anything that wants to run as both a normal middleware and an error handler has to be linked twice

s-panferov:
@reem Ok, I understand your arguments. I have thought about (2) but can't imagine such situation. Do you have some relevant example?

Anyway, current solution is comfortable enough to work with :+1:

s-panferov:
Answer to myself. Things like logging, CORS middleware or some custom header setters that must affect all responses are examples of (2).

reem:
Other open question:

Should `catch` expose a way to switch back to the normal track at all? My intuition this time said no, but I think that a "just switch back to the next middleware" might be valuable. This is how express works, since you can call `next()` with no args in an error handler to switch back to the normal track. Thoughts?

s-panferov:
@reem It makes sense in situations when you use some external middleware and want to suppress some of it's errors.

reem:
I would be far more comfortable with stuff like router raising errors on NotFound if there was a way to switch back.

reem:
You could actually still write a generic `Suppress` type which you'd just wrap the middleware whose errors you would like to ignore in.

s-panferov:
> I would be far more comfortable with stuff like router raising errors on NotFound if there was a way to switch back.

It will enable #239

reem:
Ok, I'm convinced this is desired... I'll push an update.

reem:
Done! Before I land this, I want to add some tests for the chain implementation, since it has enough indexing to make me edgy.

s-panferov:
I have read the last changes, everything seems ok for my eyes.

reem:
Ok. Tests added - 200 lines of infrastructure to allow reasonable tests...

reem:
Landing, breaking everything!

s-panferov:
:+1:

PR #278, opened 2015-01-26
Change installation instructions to use release

marrs:
Users should be using the latest release published to crates.io, rather than master.


reem:
Included in #275

PR #282, opened 2015-01-28
Fixes around std::io being renamed to std::old_io

AnthIste:
- Fixed imports and io:: references
  - Fixed deprecated Writer::write calls to use Writer::write_all


reem:
Thank you!

PR #284, opened 2015-01-29
(feat) Publicize the request module and Body.

reem:


reem:
Travis has an old rustc, so boo.

PR #285, opened 2015-01-29
(cleanup) Removed the println! that are no longer called

davidrhyswhite:
Removed the println! that are no longer called and moved a the more useful one within examples/around.rs above the listen calls.

### DIFF
--- a/examples/around.rs
+++ b/examples/around.rs
@@ -58,11 +58,11 @@ fn main() {
     let silent = Iron::new(Logger::new(LoggerMode::Silent).around(Box::new(hello_world)));
     let large = Iron::new(Logger::new(LoggerMode::Large).around(Box::new(hello_world)));
 
-    println!("Starting servers on 2000, 3000, and 4000");
     
-    tiny.listen("localhost:2000").unwrap();
-    silent.listen("localhost:3000").unwrap();
-    large.listen("localhost:4000").unwrap();
-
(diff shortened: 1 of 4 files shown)


examples/around.rs
```diff
@@ -58,10 +58,11 @@ fn main() {
     let silent = Iron::new(Logger::new(LoggerMode::Silent).around(Box::new(hello_world)));
     let large = Iron::new(Logger::new(LoggerMode::Large).around(Box::new(hello_world)));
 
+    println!("Starting servers on 2000, 3000, and 4000");
+    
     tiny.listen("localhost:2000").unwrap();
```
reem: I think this example is actually incorrect now - we'll block here and never set up `silent` or `large`.

Can you change this to `let _tiny_listening = tiny..` for each of these so that creating the server doesn't block, then bring back the println at the bottom?

davidrhyswhite:
Updated with your suggestion. Out of curiosity why does the server block after `tiny` when moving the `println!` above the server calls?

reem:
That example was actually broken before this change, moving the `println!` was unrelated. `.listen` returns an RAII guard that, when dropped, will block until the server is closed. By saving the guard into a variable using a let statement, the dropping of the guards is delayed until after they are all started (at the end of the block), removing the blocking behavior.

reem:
Builds!

reem:
Thank you very much for your work!

PR #290, opened 2015-02-11
(feat) Add ability to serve over HTTPS.

cyndis:
This adds a security property to iron::Iron, defaulting to 'Insecure'.
Setting it to 'Secure' and specifying certificate and key paths allows
the server to use HTTPS.

Also fix a FIXME in Request::from_http now that HTTPS connections
are possible.

### DIFF
--- a/src/iron.rs
+++ b/src/iron.rs
@@ -26,11 +26,38 @@ pub struct Iron<H> {
     pub handler: H,
 
     /// Once listening, the local address that this server is bound to.
-    pub addr: Option<SocketAddr>
+    pub addr: Option<SocketAddr>,
+
+    /// Once listening, the protocol used to serve content.
+    pub protocol: Option<Protocol>
+}
+
+/// Protocol used to serve content. Future versions of Iron may add new protocols
+/// to this enum. Thus you should not exhaustively match on its variants.
+pub enum Protocol {
+    /// Plaintext HTTP/1
+    Http,
+    /// HTTP/1 over SSL/TLS
+    Https {
+        /// Path to SSL certificate file
+        certificate: Path,
+        /// Path to SSL private key file
+        key: Path
+    }
+}
+
+impl Protocol {
+    /// Return the name used for this protocol in a URI's scheme part.
+    pub fn name(&self) -> &'static str {
+        match *self {
+            Protocol::Http => "http",
+            Protocol::Https { .. } => "https"
+        }
+    }
 }
 
 impl<H: Handler> Iron<H> {
-    /// Kick off the server process.
+    /// Kick off the server process using the HTTP protocol.
     ///
     /// Call this once to begin listening for requests on the server.
     /// This consumes the Iron instance, but does the listening on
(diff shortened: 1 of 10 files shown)


src/iron.rs
```diff
@@ -26,7 +26,24 @@ pub struct Iron<H> {
...
     /// Once listening, the local address that this server is bound to.
-    pub addr: Option<SocketAddr>
+    pub addr: Option<SocketAddr>,
+
+    /// Server security mode.
+    pub security: Security
+}
+
+/// Server security mode, either `Insecure` for HTTP or `Secure` for HTTPS.
+pub enum Security {
```
reem: Can this just be `Protocol` with `Http` and `Https` as variants?

src/request/mod.rs
```diff
@@ -64,7 +64,8 @@ impl<'a> Request<'a> {
     /// Create a request from an HttpRequest.
     ///
     /// This constructor consumes the HttpRequest.
-    pub fn from_http(req: HttpRequest<'a>, local_addr: SocketAddr) -> Result<Request<'a>, String> {
+    pub fn from_http(req: HttpRequest<'a>, local_addr: SocketAddr, protocol: &str)
```
reem: If we changed `Security` to `Protocol` then the `protocol` argument here could be changed to `Protocol` from `&str`.

src/iron.rs
```diff
@@ -85,8 +108,13 @@ impl<H: Handler> Iron<H> {
 
 impl<H: Handler> ::hyper::server::Handler for Iron<H> {
     fn handle(&self, http_req: HttpRequest, http_res: HttpResponse<Fresh>) {
+        let protocol = match self.security {
+            Security::Insecure => "http",
+            Security::Secure { .. } => "https"
```
reem: This should be factored into a method and called later - `Request::from_http` should take a `Protocol/Security`.

src/iron.rs
reem: Can we discourage matching on this enum exhaustively in the docs? We may later want to add other options, like `Http2`.

src/iron.rs
```diff
@@ -58,7 +75,13 @@ impl<H: Handler> Iron<H> {
         let SocketAddr { ip, port } = sock_addr.clone();
         self.addr = Some(sock_addr);
 
-        Ok(try!(Server::http(ip, port).listen_threads(self, threads)))
+        let server = match self.security {
+            Security::Insecure => Server::http(ip, port),
+            Security::Secure { ref certificate, ref key } =>
+                Server::https(ip, port, certificate.clone(), key.clone())
```
reem: It seems like it is incorrect for the `Iron` instance to own the certificate and key here. Maybe the protocol should instead be passed as an argument to `listen` and `listen_with`?

reem:
This is awesome and super exciting. I saw your work in hyper and I'm really glad to see it here too.

I just left a few comments, let me know your thoughts.

src/iron.rs
cyndis: Do you mean using two enums, one including the certificate and key and one not ("Protocol" and "ProtocolOptions" or some such)? We need to store the current protocol for use in handle(), so if we only use one enum type we end up storing the cert/key paths too.

reem: No, one enum would be used, called `Protocol`, completely equivalent to todays `Security`. The difference I am suggesting is that we pass the Protocol in through `listen` or `listen_with` rather than in `new` or some other constructor.

cyndis: Ok, sounds good.

cyndis:
Thanks :) Your suggestions are good, I'll post a v2 in a moment :)

cyndis:
Here's v2. I'll squash once the patch is ready. One thing I was wondering about was if we should add Protocol to the prelude.

reem:
How about we just provide `http(addr)` and `https(addr, certificate, key)` constructors?

cyndis:
For what? :) Iron's constructor also requires handler and Protocol doesn't contain addr.

reem:
It seems like a nuisance to have to get and type out the protocol.

cyndis:
I don't feel it would be too bad if Protocol was included in the prelude, but I could replace Iron::new with Iron::http and Iron::https taking (handler) and (handler, certificate, key) respectively. That would also closely mirror Hyper's API.

reem:
Sorry, I've been very confusing. What I meant was to replace listen with http and https.

cyndis:
Ah, ok :) That makes sense. So listen_with would still take a Protocol, just the convenience method would be replaced.

reem:
Yes.

cyndis:
v3 here.

reem:
Looks pretty good, I will read once more and merge in the morning.

cyndis:
Hi, can you merge this? :)

reem:
Ah ya, my bad for letting this spin. Just did another read through, looks like good work.

cyndis:
Thanks!

PR #291, opened 2015-02-11
(feat) Set content type automatically for based on file extension

SkylerLipthay:
Taken from iron/static#14. This pull should not be accepted as-is. `mime_types::Types::new()` loads a media type lookup table from a JSON file, and that probably shouldn't be done every request for obvious performance reasons. Any ideas as to where we can store this object? In `Iron` somehow? Maybe a static (singleton generating) method?

### DIFF
--- a/src/modifiers.rs
+++ b/src/modifiers.rs
@@ -37,7 +37,13 @@ use modifier::Modifier;
 
 use hyper::mime::Mime;
 
-use {status, headers, Request, Response, Url};
+use {status, headers, Request, Response, Set, Url};
+
+use mime_types;
+
+lazy_static! {
+    static ref MIME_TYPES: mime_types::Types = mime_types::Types::new().unwrap();
+}
 
 impl Modifier<Response> for Mime {
     #[inline]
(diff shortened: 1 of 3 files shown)


reem:
@SkylerLipthay The simplest thing is to just add another dependency on `lazy_static` and load it there.

This whole methods feels a bit not so good. It should really parse directly from the extension string to an instantiated `Mime`, which is cloned. This avoids allocating an intermediate string and the cost of re-parsing. Setting the raw header is a bit bad practice and more importantly will just move the cost of re-parsing later if it is ever accessed again.

The way we used to do it was with [rust-http-content-type](https://github.com/reem/rust-http-content-type), which generated a phf table of extension => rust-http mime instance. In this case it would be extension => hyper mime instance. Unfortunately depending on a modified version of that would ruin all hope of us being usable in 1.0 stable.

All of that said, it seems the best way forward is to do the simplest thing and use `lazy_static` to store a `Types` instance.

reem:
@SkylerLipthay needs a rebase, since I merged some other changes.

reem:
Looks good otherwise.

SkylerLipthay:
Rebased!

reem:
So everything is broken on travis right now. There's a fix on the way, so I'll just wait for it to land and merge this tomorrow.

reem:
@SkylerLipthay will need another rebase and travis broken yet again :/

SkylerLipthay:
Well, I rebased!

reem:
@SkylerLipthay finally things are actually moving, there's a small bug in the impl here, see the travis log (just a simple `expected Option<T> found T` error)

SkylerLipthay:
@reem Fixed, rebased, squashed, reworded, and macarena'd.

src/modifiers.rs
```diff
@@ -83,12 +89,12 @@ impl<'a> Modifier<Response> for &'a [u8] {
...
-    #[inline]
     fn modify(self, res: &mut Response) {
-        // Also set the content type.
-        // self.path().extension_str()
-        //     .and_then(get_content_type)
-        //     .and_then(|ct| { res.set_mut(ct) });
+        // Set the content type based on the file extension if a path is available.
+        if let Some(path) = self.path() {
+            let mime_str = MIME_TYPES.mime_for_path(path);
+            let _ = mime_str.parse().and_then(|mime: Mime| Ok(res.set_mut(mime)));
```
reem: Why not `map` here instead of `and_then`?

reem:
FINALLY

PR #293, opened 2015-02-18
Added ContentLength header in Modifier implementation for File

aleksandrpak:
Added Modifier impl for u64 to set ContentLength header. Also added ContentLength header in impl Modifier for File.

Not sure about panicking if failed to retrieve stat(). According to [this](http://man7.org/linux/man-pages/man2/stat.2.html#ERRORS) it is possible to get error even if file is valid.

Also not sure about having modifier for u64. Actually the main thing I need is to have ContentLength header for file in [iron/static](https://github.com/iron/static).

### DIFF
--- a/src/modifiers.rs
+++ b/src/modifiers.rs
@@ -62,7 +62,7 @@ impl Modifier<Response> for String {
 impl Modifier<Response> for Vec<u8> {
     #[inline]
     fn modify(self, res: &mut Response) {
-        res.headers.set(headers::ContentLength(self.len() as u64));
+        (self.len() as u64).modify(res);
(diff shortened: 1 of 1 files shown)


reem:
I'm not comfortable with this as I think it will cause confusion with many users assuming this is for setting the status code. 

The ContentLength header is available in the `iron::headers` module, which is a re-export of all the hyper headers.

aleksandrpak:
I have removed Modifier implementation for u64. Is it ok to add ContentLength header in Modifier implementation of File?

reem:
Ya, go ahead :D. Make a different PR, I'm closing this one.

PR #299, opened 2015-02-22
(chore) Enable faster travis builds.

reem:


s-panferov:
@reem could you please explain in a few words what does it do?

reem:
Enabling `sudo: false` means you are opting in to travis' container-based infrastructure, which serves up builds and has very good caches.

s-panferov:
@reem thanks

PR #303, opened 2015-03-06
Implement Modifier for &Path

panicbit:
(Note: debug builds are blocked by rust-lang/rust#22432)


reem:
Awesome, this looks great. Thank you so much for all the PRs, but as discussed before I'm gonna wait for travis to be be back up, unless you need something urgently merged.

PR #307, opened 2015-03-12
(docs): Add usage warning

s-panferov:
Let's warn people about current limitations and suggest to use a reverse proxy server.

Sorry if my language is incorrect. Feel free to say, I'll fix everything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,10 @@ FIXME: expand on when it is necessary to user hyper for serving,
 e.g. when doing HTTPS.
 -->
 
+## Usage warning
+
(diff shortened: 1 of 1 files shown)


reem:
While you might technically be right, I'm really don't *want* to put something like this in the README, it's bad for business, you know ;)

These are temporary limitations anyway, we are already thinking of better ways to distribute work over the thread pool to allow handling many more concurrent connections.

s-panferov:
@reem nobody uses a sync web-server on mid/hi-load production without nginx, ha-proxy etc. It's just a reminder for newcomers. Maybe we can just remove the sentence about limitations and leave only the recommendation?

reem:
Sure, if you craft it to be a little more positive ;)

reem:
@s-panferov you still interested in landing this?

reem:
closing due to inactivity but feel free to open another

PR #326, opened 2015-03-31
(fix) remove quotations from crate names & change hyphens to underscores

natemara:
Rustc [removed support for quotes in crate names](https://github.com/rust-lang/rust/pull/23786) and hyphens a few days ago. This PR fixes those issues.


reem:
@natemara sorry, I I just merged #325 which already includes similar changes.

PR #332, opened 2015-04-10
(doc) This addresses issue 319 to describe the threading model used by the Hyper library.

gsquire:
This is a simple documentation addition adding a few lines talking about the guard returned by a `thread::scoped` thread.

### DIFF
--- a/src/iron.rs
+++ b/src/iron.rs
@@ -64,6 +64,9 @@ impl<H: Handler> Iron<H> {
     /// This consumes the Iron instance, but does the listening on
     /// another task, so is not blocking.
     ///
+    /// The thread returns a guard that will automatically join with the parent
+    /// once it is dropped, blocking until this happens.
+    ///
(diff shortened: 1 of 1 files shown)


src/iron.rs
```diff
@@ -64,6 +64,10 @@ impl<H: Handler> Iron<H> {
     /// This consumes the Iron instance, but does the listening on
     /// another task, so is not blocking.
     ///
+    /// The threading model used does not return a thread but rather a guard.
```
reem: This is the model used by all threading in rust, so it would be better to just say "This returns a guard, which will automatically..."

reem:
Looks good, can you squash?

reem:
Also do me a favor and ping me when you do, I don't get a notification if you push new commits, so I don't know when the PR has been updated.

gsquire:
@reem I squashed the two commits and pushed it.

PR #336, opened 2015-04-21
(doc) use shields.io badges for consistency

DevoNoel:
It drives me a little nuts seeing badges with slight differences in style on a project's README.  
This PR uses [shields.io](http://shields.io) to style the Build and Crates.io badges more consistently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 Iron
 ====
 
-[![Build Status](https://img.shields.io/travis/iron/iron.svg?style=flat-square)](https://travis-ci.org/iron/iron)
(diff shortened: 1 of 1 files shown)


reem:
Sorry, but I really don't like how the crates.io badge from shields.io is orange. Maybe if you could find consistent styles with a crates.io badge that's green?

DevoNoel:
@reem Alright, well how about just using Travis' .svg badge instead?

reem:
@DevoNoel Looks great, thanks for fixing this!

heibor:
I think use mio may be a good idear, it can bing us more performance.

PR #345, opened 2015-05-16
(fix) Reexport hyper::status::StatusClass

zmbush:
Exported as iron::status::StatusClass;

Needed to support iron/logger#57


src/lib.rs
```diff
@@ -134,6 +134,8 @@ pub mod modifier {
 pub mod status {
     pub use hyper::status::StatusCode as Status;
     pub use hyper::status::StatusCode::*;
+
```
reem: can you delete this newline? otherwise LGTM

zmbush:
Fixed.

PR #346, opened 2015-05-17
(feat) allow Arc wrapped value as middleware

sunng87:
Hi, I was working on live-reload feature on my [handlebars-iron](https://github.com/sunng87/handlebars-iron/tree/feature/notify) middleware. In order to watch the file system and reload template engine, I spawned a dedicated thread waiting for fs events. The middleware is shared in main thread and watcher thread. So I made it wrapped by Arc.

I hope iron could accept Arc wrapped value as `BeforeMiddleware` and `AfterMiddleware`. Otherwise, it seems difficult to set an Arc into Chain.

The code with modified iron is here:
https://github.com/sunng87/handlebars-iron/blob/feature/notify/examples/watch_server.rs#L50

If there is a better solution for this scenario, please kindly let me know. :-)

### DIFF
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -52,6 +52,7 @@
 //!
 
 use std::any::Any;
+use std::sync::Arc;
 use {Request, Response, IronResult, IronError};
 
 /// `Handler`s are responsible for handling requests by creating Responses from Requests.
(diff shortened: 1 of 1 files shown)


reem:
Sounds good to me :)

PR #349, opened 2015-06-14
(fix) Update example in README.md compatible with version 0.1.18

AlexTalker:
But may better will be just copy-paste from examples/time.rs....

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,50 +9,40 @@ Iron
 ## Response Timer Example
 
 ```rust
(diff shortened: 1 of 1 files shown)


reem:
I think you're right and we should just copy examples/time.rs, it's easier to keep things consistent that way.

AlexTalker:
Now copy-pasted :smile:

reem:
Nice, thanks :)

PR #351, opened 2015-06-26
Hyper 0.6

AerialX:
Makes SSL a default optional feature. Removes HTTPS-related parts of the API when the feature is disabled - would a runtime panic instead be better?

Note that for the feature to actually be useful, all packages that depend on iron but do not depend on the existence of SSL will need to be updated to use `default-features = false` - so basically all middleware ever.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,11 @@ authors = [
 name = "iron"
 path = "src/lib.rs"
 
+[features]
+default = ["ssl"]
+ssl = ["hyper/ssl"]
+
 [dependencies]
-hyper = "0.5"
 typemap = "*"
 url = "*"
 plugin = "*"
@@ -31,6 +34,9 @@ conduit-mime-types = "*"
 lazy_static = "*"
 num_cpus = "*"
 
+[dependencies.hyper]
+version = "0.6"
+default-features = false
+
 [dev-dependencies]
 time = "*"
-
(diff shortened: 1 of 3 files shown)


reem:
This is great, thank you. It's a little unfortunate that making ssl optional doesn't really have an effect unless all extension crates manually opt-out. @acrichto do you have any ideas how we could use cargo to make this more ergonomic?

alexcrichton:
If the majority of dependencies don't actually need SSL support then it may want to be a legit optional feature (e.g. support not built in), but otherwise inline syntax isn't so bad:

```toml
[dependencies]
iron = { version = "...", default-features = false }
```

reem:
I'm mostly just concerned about people forgetting to set `default-features = false` even if they could. I think you're right and that might be a sign that it shouldn't be on by default.

AerialX:
I'm in favour of making it optional by default, the reason I didn't was simply to keep existing projects from breaking on the next update if they used HTTPS. In fact, I'm not really sure why Hyper didn't do so, since that update was a breaking change anyway.

The issue is that it comes down to iron having two different types of dependencies. Middleware and plugins only depend on iron to extend its functionality and need access to core types and traits. End users of the framework also want the server spawning functionality. It may make sense to split them up into two separate crates, where middleware depends on `iron_core` and then `iron` just re-exports it all for users of the framework. For now though, making HTTPS truly optional isn't bad.

Also nice inline syntax, I wasn't aware of that! FWIW, in general, a `features = ["!ssl"]` would be pretty useful, though it feels like a cheating short-circuit solution.

reem:
We can tackle the issue of optional ssl later.

Issue #311, opened 2015-03-14
More Examples


reem:
I am always on the lookout for more helpful examples of any of Iron's features.

lambdajp:
I would like to help out by creating a getting started guide as I feel this is great way to get developers up and ready quickly. The guide will provide examples and how to use the modules like urlencoded, router, etc.

reem:
@lambdajp Awesome! I'm really happy to get other people working on this too.

I'd caution against writing anything very long-form right now, since while things are relatively stable there will still be breaking changes, leading to (possibly) lots of outdated documentation. Believe me, I've had to delete hundreds of lines of documentation already.

Examples are usually much easier to update and get checked for breakage automatically by the continuous integration system. Short documentation is also easy to update, because the changes that broke it are usually near it.

lambdaburrito:
Don't worry, I believe you :) I've started to write the beginning of the guide here: https://github.com/lambdaburrito/iron/wiki/Getting-started

reem:
@lambdaburrito Cool. I'd ensure that you recommend that users read the rust book/intro first, especially since Iron uses a variety of more complex type system features.

PR #352, opened 2015-07-01
(feat) Added a new example to demonstrate Iron's router middleware.

gsquire:
This is with regards to issue [311](https://github.com/iron/iron/issues/311).


reem:
@gsquire can you remake this PR to the router repo?

reem:
We can't have it here since it would be a circular dependency, unfortunately.

Issue #353, opened 2015-07-04
Chain#link should return the chain


singpolyma:
This would allow for nice chaining syntax:

    Chain::new(handler).link(thing1).link(thing2)

AndrewBrinker:
Presumably, if `Chain.link()` is being updated, we should also update `link_before()`, `link_after()`, and `around()`, yes?

(As an aside, should `around()` be renamed to `link_around()` for consistency's sake?)

reem:
@AndrewBrinker yes and yes.

I am in favor of this change.

AndrewBrinker:
Cool. I am working on it now.

sbstp:
I think this should be closed.

PlasmaPower:
This has been fixed, I'd like to reiterate the request for this issue to be closed.

PR #355, opened 2015-07-05
(feat) Add `ChainLink` - the builder pattern for Chain

IanConnolly:
Had a similar problem to #353.

Didn't want to change the type signature on the core Chain impl, so implemented it as a Builder.

### DIFF
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -137,8 +137,19 @@ pub struct Chain {
     handler: Option<Box<Handler>>
 }
 
+/// Builder struct for a `Chain`
+///
+pub struct ChainLink {
+    befores: Vec<Box<BeforeMiddleware>>,
+    afters: Vec<Box<AfterMiddleware>>,
+
+    // Internal invariant: this is always Some
+    // Invariant inherited from the associated `Chain`
+    handler: Option<Box<Handler>>
+}
+
 impl Chain {
-    /// Construct a new ChainBuilder from a `Handler`.
+    /// Construct a new Chain from a `Handler`.
     pub fn new<H: Handler>(handler: H) -> Chain {
         Chain {
             befores: vec![],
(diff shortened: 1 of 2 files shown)


IanConnolly:
Realised I hadn't included an example anywhere!

The example in the README: 

```rust
fn main() {
    let mut chain = Chain::new(hello_world);
    chain.link_before(ResponseTime);
    chain.link_after(ResponseTime);
    Iron::new(chain).http("localhost:3000").unwrap();
}
```

becomes:

```rust
fn main() {
    let chain = ChainLink::new(hello_world)
        .link_before(ResponseTime)
        .link_after(ResponseTime)
        .lock_chain();
    Iron::new(chain).http("localhost:3000").unwrap();
}
```

reem:
I agree we want something like this, but I have to think about the API a bit.

Regardless, one major comment on the implementation ChainLink should just contain a Chain and call its methods rather than reimplementing the internals.

IanConnolly:
Good spot, will push that fix this evening.

I'm using this patch on a project myself, so it'd be great to get something like this into Iron. I'm happy to put the work in to get it to happen.

If I end up getting to go to the meetup at Mozilla tomorrow I'll try and grab you to chat about it.

src/middleware/mod.rs
```diff
@@ -178,6 +178,60 @@ impl Chain {
     }
 }
 
+/// Builder struct for a `Chain`
+///
+pub struct ChainLink {
```
nixpulvis: I'd probably use a wrapper struct like `ChainLink(Chain)`, because the name of the field of your struct doesn't really add any information.

nixpulvis:
Seems like most of the implementations could be abstracted with a macro.

posix4e:
Should we close this?

Hoverbear:
I'm going to close this PR since it's over a year old now. 

Thank you @IanConnolly for submitting this PR. :)

PR #356, opened 2015-07-10
Rename "relative scheme" to "special scheme" according to URL spec.

### LINKED ISSUE #309: Better Documentation for "relative" Urls
In iron/src/request/url.rs, it says about Url::parse(input: &str) that "The input must be a valid URL in a relative scheme for this to succeed."

However, valid relative URLs cause a panic. The upstream repo, rust-url docstring says: "In particular, relative URL references are parse errors since no base URL is provided."

So there's a mismatch there. Is the documentation here wrong, or has the upstream API changed...?

diogovk:
See: https://lists.w3.org/Archives/Public/public-whatwg-archive/2015Jun/0028
https://github.com/whatwg/url/commit/d5470b8d02

Closes #309

### DIFF
--- a/src/request/url.rs
+++ b/src/request/url.rs
@@ -54,11 +54,11 @@ pub struct Url {
 impl Url {
     /// Create a URL from a string.
     ///
-    /// The input must be a valid URL in a relative scheme for this to succeed.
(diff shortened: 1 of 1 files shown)


reem:
Makes sense to me. Thanks for this!

PR #358, opened 2015-07-13
(refactor) iron.rs

leodasvacas:
Eliminates use of the `return` keyword. Clearer `fn bad_request`. Refactored `fn handle` into a single expression.


reem:
LGTM, thanks!

PR #361, opened 2015-07-22
minor changes

ghost:


.travis.yml
```diff
@@ -1,4 +1,6 @@
 language: rust
+rust:
```
reem: can you add both stable and nightly instead of hardcoding a version?

netmml: @reem added nightly, 1.1.0 has to be defined for the stable part though .. beta is available too if wanted

reem: Great, beta is not needed. Can you squash these changes and I will merge?

netmml: @reem I removed 1.1.0 since travis downloads the latest stable rust release by itself, but nightly has added and requested changes have been made

reem: Hmm, looking at the travis log it appears only nightly is tested now. I think having `1.1.0` is necessary.

netmml: @reem I added stable instead which is the equivalent to 1.1.0

reem:
@netmml ok, looks good to me now. Can you squash?

netmml:
@reem sorry, using github's web interface

reem:
Squashed manually in #379

PR #362, opened 2015-07-23
(fix) Don't reject HTTP <= 1.0 requests without a Host header.

### LINKED ISSUE #327: Bad request when no host is specified
I'm dealing with an HTTP client that is sending a request with no `Host` header and so I'm getting [this error](https://github.com/iron/iron/blob/fd5ccecc4ed7d5a79bcf8654848f51ce188756a8/src/request/mod.rs#L88):
```
ERROR:iron::iron: Error creating request:
    No host specified in request
```
The comment in the code says it's only required for HTTP/1.1, but this particular client is using HTTP/1.0. Would it be possible to do a version check here and only fail here if there is no host and the version is HTTP/1.1?

This may seem like a strange case, but this is for an application that isn't a website/web server but still uses the HTTP protocol. Using `hyper` directly might be a better fit, but this still seems like it could be fixed here.

ahwatts:
This sort of fixes #327 .

This fix gets me past the issue I'm seeing, but it's not ideal. I'm assuming the local_addr is good enough to generate a URL against, but that's probably not going to be true often (ever?). I'm open to better suggestions.

### DIFF
--- a/src/request/mod.rs
+++ b/src/request/mod.rs
@@ -7,6 +7,7 @@ use std::fmt::{self, Debug};
 use hyper::uri::RequestUri::{AbsoluteUri, AbsolutePath};
 use hyper::net::NetworkStream;
 use hyper::http::h1::HttpReader;
+use hyper::version::HttpVersion;
 
 use typemap::TypeMap;
 use plugin::Extensible;
@@ -68,7 +69,7 @@ impl<'a, 'b> Request<'a, 'b> {
     /// This constructor consumes the HttpRequest.
     pub fn from_http(req: HttpRequest<'a, 'b>, local_addr: SocketAddr, protocol: &Protocol)
                      -> Result<Request<'a, 'b>, String> {
-        let (addr, method, headers, uri, _, reader) = req.deconstruct();
+        let (addr, method, headers, uri, version, reader) = req.deconstruct();
 
         let url = match uri {
             AbsoluteUri(ref url) => {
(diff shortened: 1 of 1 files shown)


src/request/mod.rs
```diff
@@ -79,13 +80,21 @@ impl<'a, 'b> Request<'a, 'b> {
...
+                let url_string = match (version, headers.get::<headers::Host>()) {
+                    (_, Some(ref host)) => {
+                        // Attempt to prepend the Host header (mandatory in HTTP/1.1)
+                        format!("{}://{}:{}{}", protocol.name(), host.hostname, local_addr.port(), path)
                     },
-                    None => return Err("No host specified in request".to_string())
+                    (v, None) if v < HttpVersion::Http11 => {
+                        // Attempt to use the local address? (host header is not required in HTTP/1.0).
+                        match local_addr {
+                            SocketAddr::V4(addr4) => format!("{}://{}:{}{}", protocol.name(), addr4.ip(), local_addr.port(), path),
```
untitaker: I'm kind of unhappy to write addr serialization code here. Isn't there something in `url` that we can use for this?

reem:
Closing due to inactivity but feel free to reopen with the requested changes.

src/request/mod.rs
Hoverbear: This is what was already used so there is no reason for this to block this PR.

PR #363, opened 2015-07-24
Distinguish error by using status::BadRequest

mneumann:
Clarify the error example by using a status::BadRequest in the error
case, and status::Ok in the normal case (which is never triggered).


reem:
Great! I like the comment too, helps clarify what's going on.

PR #366, opened 2015-08-02
Switch response abstraction from `Read` to `WriteBody`.

sfackler:
A `Read`er is not the best abstraction for the response body - it has to
be copied to the output `Write`r through an extra intermediary buffer
and complicates the implementation for users that want to compute the
response on the fly.

This design is modeled after Java JAX-RS's StreamingOutput class. The
`write_body` method takes the Hyper `Response` object concretely instead
of a generic `Write`r to allow for future optimizations with Hyper, for
example, allowing the underlying file descriptor to be retrieved for use
with `sendfile` on Linux.

This is a [breaking-change]

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -1,19 +1,63 @@
 //! Iron's HTTP Response representation and associated methods.
 
-use std::io::{self, Read, Write};
+use std::io::{self, Write};
 use std::fmt::{self, Debug};
+use std::fs::File;
 
 use typemap::TypeMap;
 use plugin::Extensible;
-use modifier::{Set, Modifier};
-
+use modifier::{Set, Modifier}; 
 use hyper::header::Headers;
 
 use status::{self, Status};
 use {Plugin, headers};
 
 pub use hyper::server::response::Response as HttpResponse;
-use hyper::net::Fresh;
+use hyper::net::{Fresh, Streaming};
+
+/// A trait which writes the body of an HTTP response.
+pub trait WriteBody {
+    /// Writes the body to the provided `Response<Streaming>`.
+    fn write_body(&mut self, res: &mut HttpResponse<Streaming>) -> io::Result<()>;
+}
+
+impl WriteBody for String {
+    fn write_body(&mut self, res: &mut HttpResponse<Streaming>) -> io::Result<()> {
+        self.as_bytes().write_body(res)
+    }
+}
+
+impl<'a> WriteBody for &'a str {
+    fn write_body(&mut self, res: &mut HttpResponse<Streaming>) -> io::Result<()> {
+        self.as_bytes().write_body(res)
+    }
+}
+
+impl WriteBody for Vec<u8> {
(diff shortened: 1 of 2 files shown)


sunng87:
I didn't look into the code but I'm +1 for this idea. Most Ring-like framework has similar issue. 

When I work on Handlebars, I hope Iron response could give me a `Write` instead of `Read`. So I won't have to generate pages into a string, then copy it into response.

src/response.rs
```diff
@@ -1,19 +1,63 @@
...
-
+use modifier::{Set, Modifier}; 
 use hyper::header::Headers;
 
 use status::{self, Status};
 use {Plugin, headers};
 
 pub use hyper::server::response::Response as HttpResponse;
-use hyper::net::Fresh;
+use hyper::net::{Fresh, Streaming};
```
reem: can you `pub use` `Streaming` so implementors don't have to depend on hyper explicitly to implement `WriteBody`.

reem:
As discussed earlier I am definitely for this. I'm hesitating about exposing the hyper response directly, but I think it's probably the right thing to do here.

sfackler:
I'd be totally on board with defining an opaque wrapper around the hyper response but think it still makes sense to have it be a concrete type instead of generic over `Write`.

sfackler:
Updated with the wrapper type.

mneumann:
Would be nice to see a small example how this can be used to write out a response from a regular iron handler.

bfrog:
If this means I can write my handler like...

``` rust
fn handle_req(req: Request, res: Response) ->
    res.WriteHeader();
    for x in y {
         res.Write(...);
    }
    // do more stuff after writting out response, like write some stats to a db somewhere, log request     somewhere, etc
}
```

That would be fantastic, its the main reason I'm not using iron at the moment, there's no clear way to do things after writing out a response.

sfackler:
@mneumann Here's one example of generating the response while sending it:

```rust
struct WriteNumbers;

impl WriteBody for WriteNumbers {
    fn write_body(&mut self, body: &mut ResponseBody) -> IronResult<()> {
        for i in 0..10000 {
            try!(write!(body, "{}", i));
        }
    }
}

fn handle_write_numbers(_: &mut Request) -> IronResult<Response> {
    let body: Box<WriteBody + Send> = Box::new(WriteNumbers);
    Ok(Response::with(body))
}
```

@bfrog That's not really how this API is structured. That seems like a job for [AfterMiddleware](http://ironframework.io/doc/iron/middleware/trait.AfterMiddleware.html)?

reem:
@bfrog can you actually open an issue for that feature request? It makes perfect sense you'd want to run actions post-actual-response-writing but the current API does not really allow that.

@sfackler the reason it can't be an AfterMiddleware is because those still run before the response is actually written out.

Sorry for taking so long to merge this, I've just been thinking about it and I also want to ensure I merge it when I have time to go and update the other repos in `iron`.

Issue #373, opened 2015-08-17
No way to use `addr` field of `Iron` struct?


cmbrandenburg:
How are we supposed to use the `addr` field of the `Iron` struct in a meaningful way? The field is populated with the server's bound-to address _after_ listening begins, but listening begins by calling the `http`, `https`, or `listen_with` functions—all of which consume the `Iron` instance.

Am I missing something?

reem:
It is used to populate the `local_addr` field of `Request` in `Request::from_http`, and can be "accessed" through that. It is meant to be an implementation detail of the `Iron` type.

PR #374, opened 2015-08-17
Mark Iron struct fields as 'internal' in doc

cmbrandenburg:
I hope no one else will spend a few minutes puzzling over this one as I did.

(See issue #373.)

### DIFF
--- a/src/iron.rs
+++ b/src/iron.rs
@@ -26,10 +26,10 @@ pub struct Iron<H> {
     /// requests.
     pub handler: H,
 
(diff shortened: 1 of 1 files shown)


reem:
Hmm, can you instead just mark them `#[doc(hidden)]`?

reem:
Hmm, it might actually be better to just make them private fields.

untitaker:
I made those fields private in https://github.com/iron/iron/commit/e88a4f3a915ed726ae3a45ecefbac3952afda421

PR #375, opened 2015-08-18
Manage before middleware error

ilkosta:
Hi, 

I added 

* a message to help the user running the example
* an around middleware to trace on the console the error message catched by the server

### DIFF
--- a/examples/error.rs
+++ b/examples/error.rs
@@ -2,7 +2,7 @@ extern crate iron;
 extern crate time;
 
 use iron::prelude::*;
-use iron::{Handler, BeforeMiddleware};
+use iron::{Handler, BeforeMiddleware, AroundMiddleware};
 use iron::status;
 
 use std::error::Error;
@@ -37,19 +37,63 @@ impl Handler for ErrorHandler {
 
 impl BeforeMiddleware for ErrorProducer {
     fn before(&self, _: &mut Request) -> IronResult<()> {
-        Err(IronError::new(StringError("Error".to_string()), status::BadRequest))
+        Err(IronError::new(StringError("Error you cannot access the resource!".to_string()), status::BadRequest))
     }
 }
 
+
+// to handle the Error message by using an AroundMiddleware
+struct LogEnabler;
+
+struct HandlerWithLog<H: Handler> {
+    handler: H
+}
+
+impl AroundMiddleware for LogEnabler {
+
+    fn around(self, handler: Box<Handler>) -> Box<Handler> {
+        Box::new(HandlerWithLog { handler: handler } ) as Box<Handler>
+    }
+}
+ 
+impl <H: Handler> Handler for HandlerWithLog<H> {
+
+    fn handle(&self, req: &mut Request) -> IronResult<Response> {
+    
+        let res = self.handler.handle(req);
+        
+        {
+          match res {
+            Ok(_) => panic!("is not possible!"),
+            Err(IronError {error: ref what_went_wrong, ref response }) => println!("what went wrong: {:?} - response: {:?}", what_went_wrong, response)
+          };          
+        }
+        
+        res
+    }
+}
+
+pub fn get_log_enabled_handler(handler : Box<Handler>) -> Box<Handler> {
+  
+  let use_log = LogEnabler;
+    
+  use_log.around(handler)
+}
+
+
 fn main() {
     // Handler is attached here.
+    
     let mut chain = Chain::new(ErrorHandler);
 
     // Link our error maker.
     chain.link_before(ErrorProducer);
-
+    
+    // enable the tracing of the error by using an AroundMiddleware
+    let logged_handler = get_log_enabled_handler(Box::new(chain));
+    
     println!("server running at http://localhost:3000");
     println!("try it by opening the url in a browser or by: 'curl -v http://localhost:3000', 'curl -v http://localhost:3000/any_other_destination/same_bad_request'");
-    Iron::new(chain).http("localhost:3000").unwrap();
+    Iron::new(logged_handler).http("localhost:3000").unwrap();
 }
 
--- a/examples/error.rs
+++ b/examples/error.rs
@@ -48,6 +48,8 @@ fn main() {
     // Link our error maker.
     chain.link_before(ErrorProducer);
 
+    println!("server running at http://localhost:3000");
+    println!("try it by opening the url in a browser or by: 'curl -v http://localhost:3000', 'curl -v http://localhost:3000/any_other_destination/same_bad_request'");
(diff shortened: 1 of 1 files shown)


examples/error.rs
```diff
@@ -37,17 +37,63 @@ impl Handler for ErrorHandler {
...
+
+// to handle the Error message by using an AroundMiddleware
+struct LogEnabler;
+
+struct HandlerWithLog<H: Handler> {
+    handler: H
+}
+
+impl AroundMiddleware for LogEnabler {
+
```
reem: delete this newline

examples/error.rs
```diff
@@ -37,17 +37,63 @@ impl Handler for ErrorHandler {
...
+    }
+}
+ 
+impl <H: Handler> Handler for HandlerWithLog<H> {
+
+    fn handle(&self, req: &mut Request) -> IronResult<Response> {
+    
+        let res = self.handler.handle(req);
+        
+        {
```
reem: why does this need to be in its own block?

examples/error.rs
```diff
@@ -37,17 +37,63 @@ impl Handler for ErrorHandler {
...
+          match res {
+            Ok(_) => panic!("is not possible!"),
+            Err(IronError {error: ref what_went_wrong, ref response }) => println!("what went wrong: {:?} - response: {:?}", what_went_wrong, response)
```
reem: if you just do `IronError { ref error, ref response }` I think that will be clearer

examples/error.rs
```diff
@@ -37,17 +37,63 @@ impl Handler for ErrorHandler {
...
+impl AroundMiddleware for LogEnabler {
+
+    fn around(self, handler: Box<Handler>) -> Box<Handler> {
+        Box::new(HandlerWithLog { handler: handler } ) as Box<Handler>
+    }
+}
+ 
+impl <H: Handler> Handler for HandlerWithLog<H> {
+
+    fn handle(&self, req: &mut Request) -> IronResult<Response> {
```
reem: I think the best way to write this method would be something like:

```rust
self.handler.handle(req).map_err(|e| println!(..); e)
```

examples/error.rs
```diff
@@ -37,17 +37,63 @@ impl Handler for ErrorHandler {
...
+    
     let mut chain = Chain::new(ErrorHandler);
 
     // Link our error maker.
     chain.link_before(ErrorProducer);
-
-    Iron::new(chain).http("localhost:3000").unwrap();
+    
+    // enable the tracing of the error by using an AroundMiddleware
+    let logged_handler = get_log_enabled_handler(Box::new(chain));
```
reem: why not just `logged_handler = LogEnable.around(Box::new(chain))`

examples/error.rs
```diff
@@ -37,17 +37,63 @@ impl Handler for ErrorHandler {
...
+    
+    println!("server running at http://localhost:3000");
+    println!("try it by opening the url in a browser or by: 'curl -v http://localhost:3000', 'curl -v http://localhost:3000/any_other_destination/same_bad_request'");
```
reem: better as documentation/comments then prints I think

reem:
I left some specific comments, but at a higher level you actually do not need to use an AroundMiddleware here, an AfterMiddleware would do perfectly fine, since all you need to do is introspect the response.

ilkosta:
Really thanks! 
This is better than I ever hoped.
As you have guessed I'm trying to learn rust using an Iron project as gym, and my pull is for a review :)

I missed completely `map_err` as well as the method invocation of a struct. I thought that the structs went always instantiated and binded for the `ref.method() syntax`.

Now I see that an AroundMiddleware was not necessary. Apart from this example I am interested in the use case of a rest webservice that uses json and wants to "describe" errors for not met requirements. In that case, with an AfterMiddleweare I cannot change the response content, only the response headers. True?

reem:
@ilkosta actually you can change the response content, or even replace the response wholesale - nothing is written to the actual stream until the top-level Handler has returned (which necessitates all middleware having run).

examples/error.rs
```diff
@@ -37,17 +36,32 @@ impl Handler for ErrorHandler {
 
 impl BeforeMiddleware for ErrorProducer {
     fn before(&self, _: &mut Request) -> IronResult<()> {
-        Err(IronError::new(StringError("Error".to_string()), status::BadRequest))
+        Err(IronError::new(StringError("You cannot access the resource!".to_string()), status::BadRequest))
+    }
+}
+
+// the error is simply printed on stdout and returned as is
+struct ErrorHandler;
```
reem: add a newline after this declaration

examples/error.rs
```diff
@@ -37,17 +36,32 @@ impl Handler for ErrorHandler {
 
 impl BeforeMiddleware for ErrorProducer {
     fn before(&self, _: &mut Request) -> IronResult<()> {
-        Err(IronError::new(StringError("Error".to_string()), status::BadRequest))
+        Err(IronError::new(StringError("You cannot access the resource!".to_string()), status::BadRequest))
+    }
+}
+
+// the error is simply printed on stdout and returned as is
```
reem: Generally speaking can you make new comments full sentences with capital letters, punctuation, etc.? It makes the examples easier to read for future new users.

examples/error.rs
```diff
@@ -37,17 +36,32 @@ impl Handler for ErrorHandler {
...
+    let mut chain = Chain::new(NeverCalledHandler);
 
     // Link our error maker.
     chain.link_before(ErrorProducer);
 
+    // enable the tracing of the error by catching it with an AfterMiddleware
+    chain.link_after(ErrorHandler);
+
+    println!("server running at http://localhost:3000");
+    println!("try it by opening the url in a browser or by: 'curl -v http://localhost:3000', 'curl -v http://localhost:3000/any_other_destination/same_bad_request'");
```
reem: I would rather these prints were replaced by comments.

examples/error.rs
```diff
@@ -37,17 +36,32 @@ impl Handler for ErrorHandler {
...
-        Err(IronError::new(StringError("Error".to_string()), status::BadRequest))
+        Err(IronError::new(StringError("You cannot access the resource!".to_string()), status::BadRequest))
+    }
+}
+
+// the error is simply printed on stdout and returned as is
+struct ErrorHandler;
+impl AfterMiddleware for ErrorHandler {
+    fn catch(&self, _: &mut Request, err: IronError) -> IronResult<Response> {
+      println!("handled the error: {:?}",err.error);
```
reem: How about `println!("Request Error! Error was: {:?}", err.error)`?

reem:
Closing due to inactivity but feel free to reopen with the requested changes.

PR #378, opened 2015-08-21
(chore) Pin versions of all dependencies to their current minor release.

reem:


reem:
cc @sfackler no more *s!

sfackler:
:confetti_ball:

Cargo.toml
```diff
@@ -24,15 +24,15 @@ default = ["ssl"]
...
-plugin = "*"
-modifier = "*"
-error = "*"
-log = "*"
-conduit-mime-types = "*"
-lazy_static = "*"
-num_cpus = "*"
+typemap = "0.3"
+url = "0.2"
+plugin = "0.2*"
```
sfackler: Looks like a trailing wildcard here

reem: Oh woops, what behavior does this actually have?

sfackler: No idea. Seems like it shouldn't parse.

reem: Surprisingly enough it does work...

PR #380, opened 2015-08-28
Fix prelude doc formatting.

steveklabnik:
Previously, the 'for ' part blended into the description on the main page.
By putting it on a new line, the index is better formatted.


reem:
Thanks :)

PR #381, opened 2015-08-28
Make the hello world a real hello world.

steveklabnik:
I had to figure out what all to import to actually get this to work.

No worries if you don't want to take this, but I think showing the full thing is much better in this case.


reem:
It does make the example longer, but also more useful I agree.

PR #389, opened 2015-10-06
Adding examples

pdib:
Hi,
I've started using Iron for personal projects and have written some example code while learning to use the framework. I added two examples:

- `error_recovery.rs` illustrates how to recover from an error in a middleware piece. I tried that to see how the request switches from the normal flow to the error flow.

- `simple_routing.rs` illustrates how to select a specific handler depending on the request. (I learned about the [routing middleware](https://github.com/iron/router) afterwards, which is obviously a better solution, but I think the example can be useful for learning.)

I hope this can help people discovering the framework.

### DIFF
--- a/examples/error_recovery.rs
+++ b/examples/error_recovery.rs
@@ -1,15 +1,16 @@
-// This example illustrates the error flow of a Request in BeforeMiddleware.
+// This example illustrates the error flow of a Request in the middleware Chain.
 // Here is the chain used and the path of the request through the middleware pieces:
 //
-// Normal Flow : ____[ErrorProducer::before]_____     [ErrorRecover::before]     ____[HelloWorldHandler::handle]___
-// Error Flow  :     [ErrorProducer::catch ]     |____[ErrorRecover::catch ]_____|
-
+// Normal Flow : __[ErrorProducer::before]__   [ErrorRecover::before]  __[handle::HelloWorldHandler]__[ErrorProducer::after]__   [ErrorRecover::after]  __ ...
+// Error Flow  :   [ErrorProducer::catch ]  |__[ErrorRecover::catch ]__|                              [ErrorProducer::catch]  |__[ErrorRecover::catch]__|
+//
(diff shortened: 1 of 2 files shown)


reem:
Awesome! This is exactly the kind of thing we need to make iron more approachable.

I've given the two examples a read and I have two main thoughts:
   - the error recovery example would be even better if it also included AfterMiddleware, to make it clear what happens in that case (even if the behavior is basically exactly the same)
   - the router example should mention the real `router` crate/handler just so people know it exists

examples/error_recovery.rs
```diff
@@ -0,0 +1,67 @@
...
+        Ok(Response::with((status::Ok, "Hello world !")))
+    }
+}
+
+impl BeforeMiddleware for ErrorProducer {
+    fn before(&self, _: &mut Request) -> IronResult<()> {
+        // The error produced here switches to the error flow.
+        // The catch method of following middleware pieces will be called.
+        // The Handler will be skipped unless the error is handled by another middleware piece.
+        Err(IronError::new(StringError("Error in ErrorProducer".to_string()), status::BadRequest))
```
reem: I would also mention here that the `status::BadRequest` is acting as a modifier and that `IronError` contains a `Response` which will be sent if the error is not otherwise handled.

examples/error_recovery.rs
```diff
@@ -0,0 +1,67 @@
...
+    }
+}
+
+impl BeforeMiddleware for ErrorRecover {
+    fn catch(&self, _: &mut Request, err: IronError) -> IronResult<()> {
+        // We can use the IronError from previous middleware to decide what to do.
+        // Returning Ok() from a catch method resumes the normal flow.
```
reem: Would be good to note here that this means the Handler will be run, since we entered back into the normal flow before the Hander.

pdib:
Thanks for the feedback, I updated the PR.
Also, tell me if you have other topics in mind that could be turned into helpful examples.

untitaker:
Excellent, thanks!

PR #401, opened 2015-10-27
(refactor) Added return to link functions on Chain

AndrewBrinker:
This commit modifies `Chain.link()`, `Chain.link_before()`, `Chain.link_after()`, and `Chain.around()` to return a reference to the current `Chain`. This allows the, ahem, "chaining" of calls to these functions, which is slightly more ergonomic than the way things are now.

All tests pass locally with these changes, but I'm not sure how to best add tests to check this specific change.

This addresses issue #353.

### DIFF
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -151,30 +151,37 @@ impl Chain {
     ///
     /// Middleware that have a Before and After piece should have a constructor
     /// which returns both as a tuple, so it can be passed directly to link.
-    pub fn link<B, A>(&mut self, link: (B, A))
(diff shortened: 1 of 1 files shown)


reem:
You probably meant to return `&mut Chain`, not `&Chain`. The calls cannot actually be chained on a `&Chain`, which is immutable.

AndrewBrinker:
Oh geez, you're right. Fixing.

reem:
This seems like a pure win to me, it isn't even too bad of a breaking change.

reem:
Can you squash and I will merge?

AndrewBrinker:
Squashed. :smile:

reem:
Awesome!

PR #402, opened 2015-10-27
(refactor) Rename around() to link_around()

AndrewBrinker:
This is a small change that just renames `Chain.around()` to `Chain.link_around()`, for consistency with the other linking functions.

### DIFF
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -176,7 +176,16 @@ impl Chain {
     }
 
     /// Apply an `AroundMiddleware` to the `Handler` in this `Chain`.
+    ///
+    /// Note: This function is being renamed `link_around()`, and will
+    /// eventually be removed.
     pub fn around<A>(&mut self, around: A) -> &mut Chain
+    where A: AroundMiddleware {
+        self.link_around(around)
+    }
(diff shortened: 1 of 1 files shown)


reem:
This has a conflict with the last PR you made, which I just merged. I am hesitant to land this immediately as it is a breaking change. Could you instead make a new function link_around, and we can remove the old around function later?

AndrewBrinker:
Yup. Can do. I knew they would conflict, but figured making two commits was
good for history-tracking purposes.

AndrewBrinker:
Okay, I've rebased on top of master, and added back the `around()` function with an additional comment about deprecation.

reem:
Given that we plan to deprecate around, the real implementation should live int he body of link_around and be called from around, rather than the other way around.

AndrewBrinker:
Yup. Fixed.

untitaker:
Are there any appropriate markers that we can apply to the deprecated function?

reem:
Not currently. We will likely get `#[deprecated]` back in the near future though.

reem:
Merged anyway, we can deal with removing `around` when the time comes.

PR #403, opened 2015-10-30
(feat) Add helper macros for error handling

untitaker:


reem:
Can you add a short note to the docs that these macros are experimental? Otherwise LGTM.

untitaker:
Done, also I changed the default modifier of some_or.

untitaker:
Ok, **now** I'm done.

The idea here is to use `itry` for internal operations, and `some_or` for parsing the request.

* `itry!(write_to_db())`
* `some_or!(get_a_url_param())`

Since this is not really a hard requirement in any way, I'm not sure whether to include this into the docs.

reem:
I'm not totally sure how to feel about the different modifiers, but I think if we are going to go that way we should include a note in the docs with the suggested use cases. I think it would be fine to merge the macros themselves as is now, we can always iterate on them if they aren't as convenient as imagined for real use cases.

untitaker:
Added a note.

untitaker:
Also I think I'm going to rework the examples...

untitaker:
there, I also updated the examples. I didn't find a good usecase for `some_or`, to build a better example I'd have to pull in `urlencoded`.

untitaker:
We need a better name for `some_or`. Perhaps `iexpect`?

reem:
I like `iexpect`.

untitaker:
renamed to `iexpect`.

examples/helper_macros.rs
```diff
@@ -0,0 +1,27 @@
...
+
+use iron::prelude::*;
+use iron::status;
+use iron::method;
+
+fn main() {
+    Iron::new(|req: &mut Request| {
+        Ok(match req.method {
+            method::Get => {
+                let f = itry!(fs::File::open("foo.txt"), (status::Ok, ""));
```
reem: I think it would be better to respond with status::NotFound if the file doesn't exist

examples/helper_macros.rs
```diff
@@ -0,0 +1,27 @@
...
+                Response::with(status::Ok).set(f)
```
reem: this can/should be written `Response::with((status::Ok, f))`

reem: possibly we should change the impl of Modifier for Path to send `NotFound` if the file doesn't exist, so this whole thing could be `Response::with((status::Ok, Path::new("foo.txt")))

src/macros.rs
```diff
@@ -0,0 +1,35 @@
...
+/// ```
+///
+#[macro_export]
+macro_rules! itry {
+    ($result:expr) => (itry!($result, ::iron::status::InternalServerError));
+
+    ($result:expr, $modifier:expr) => (match $result {
+        ::std::result::Result::Ok(val) => val,
+        ::std::result::Result::Err(err) => return ::std::result::Result::Err(
+            ::iron::IronError::new(err, $modifier))
```
reem: this should be written `$crate::IronError`..., same below

examples/helper_macros.rs
untitaker: IMO that's too implicit behavior. If I pass a path I probably assume it exists.

untitaker: Honestly I'm a bit surprised that Path impls Modifier, it seems like a unnecessary specialization of passing a File to the response.

examples/helper_macros.rs
untitaker: It doesn't really matter but I wanted the example to have a default value.

examples/helper_macros.rs
reem: It exists now so that if you pass a Path the content-type will be set automatically.

untitaker:
@reem any update?

untitaker:
\o/

PR #411, opened 2015-12-07
(fix) Add re-export of hyper::server::Listening [Fixes #406]

### LINKED ISSUE #406: Re-export Hyper::server::Listening
Is there a re-export for `Hyper::Server::Listening` somewhere (the object returned by `Iron::http`)? Or do I have to add hyper as an explicit dependency, just because I want to be able to stop a server I started?

If not, it would be nice if you could add the re-export :) In my opinion all the API of Iron should be importable from Iron :)

rnestler:
This re-exports `hyper::server::Listening` as `iron::Listening`. Since iron returns an `Listening` in `Iron::http()` and `Iron::https()` it should re-export it to make it available for client code.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -82,7 +82,7 @@ pub use middleware::{BeforeMiddleware, AfterMiddleware, AroundMiddleware,
                      Handler, Chain};
 
 // Server
-pub use iron::{Iron, Protocol};
+pub use iron::{Iron, Protocol, Listening};
 
 // Extensions
 pub use typemap::TypeMap;


untitaker:
Not sure why we don't export everything from `iron::iron`. @reem?

dbrgn:
Would be great if this could be handled. Using the hyper crate explicitly just to access `Listening` or other types from hyper is unnecessary and can cause version conflicts.

reem:
Superseded by #431 which re-exports all of the iron module.

Issue #386, opened 2015-09-30
Allow the user to provide a bound TcpListener to the server


4xrsJCr9:
This would support the scenario in which a process (running as root) binds a privileged port and then `setuid's` to an unprivileged user.

Current usage leaves a small window (theoretically, at least) between starting the server and calling `setuid` in which requests could be processed as root.

reem:
This would definitely be useful. I've started work towards solving this in the above linked PR to hyper.

sfackler:
This is implemented in #510

PR #415, opened 2015-12-29
Create using custom listener

viraptor:
Adding a way to start the server using a custom listener. This allows starting
up without explicitly opening sockets, which enables use cases like systemd
socket activation. (separate crate will be released soon, already works with
this interface)

### DIFF
--- a/src/iron.rs
+++ b/src/iron.rs
@@ -9,6 +9,7 @@ use std::path::PathBuf;
 pub use hyper::server::Listening;
 use hyper::server::Server;
 use hyper::net::Fresh;
+use hyper::net::NetworkListener;
 
 use request::HttpRequest;
 use response::HttpResponse;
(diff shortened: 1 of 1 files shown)


erikjohnston:
I want to be able to do something like this so that I can use a customized `SslContext`.

An alternative would be to allow starting iron with a completely preconfigured `hyper::server::Server`. As far as I can see, iron only needs two things: the name of the scheme used (i.e. http or https) and the local addr. Ideally we would be able to get both from the `hyper::server::Server`, but those don't seem to be currently exposed.

zofrex:
I'm really keen to see a fix for #386 to land as well. I've also made a potential patch to resolve it but I'm not in love with my solution.

Regarding the failed build above, I reproduced this locally but the error went away once I rebased onto master, so it's good to go on that front.

To aid in thinking about the interface (and perhaps document it if this gets merged), I've made a minimal example of using it:

(I'm not 100% certain I'm doing this the easiest way possible so let me know if I'm missing something!)

```
extern crate iron;
extern crate hyper;

use iron::prelude::*;
use iron::status;
use iron::Protocol;
use std::net::TcpListener;
use hyper::net::HttpListener;

fn main() {
    fn hello_world(_: &mut Request) -> IronResult<Response> {
        Ok(Response::with((status::Ok, String::from("Hello World!"))))
    }

    let listener = TcpListener::bind("0.0.0.0:80").unwrap();

    let _i = Iron::new(hello_world)
                 .listen_custom(None, HttpListener::from(listener), Protocol::Http, None)
                 .unwrap();
}
```

I really like the use of Option for the thread count (which might help reduce some of the repetition in this area if applied more broadly?). This is waaay simpler and more elegant a change than my own attempt at this, too. One thing - I'm not super thrilled with exposing hyper's implementation details this way, as it means I have to add hyper to my Cargo.toml and make sure to specify the same version hyper uses (is there a workaround for this?)

Hoverbear:
Is there still interest in this PR? Is it still relevant?

zofrex:
@Hoverbear I've upgraded to the latest Iron (0.4.0) and I don't see a way to accomplish my goals (bind to socket, drop privs, start Iron) without either this patch or #457. Currently I am maintaining my own fork of Iron and overriding the dependency in my project to do this.

So yes, as far as I know, this is still relevant – and an emphatic yes for interest :)

sfackler:

untitaker:
Nice @sfackler!

PR #418, opened 2016-01-18
Update doc about default number of threads

iwinux:
To make it consistent with what the [doc](https://github.com/iwinux/iron/blob/165daff62dfc6c0050f6706633c7a3f08b7b8dfe/src/iron.rs#L105) says:

> Defaults to a threadpool of size `2 * num_cpus`.


iwinux:
oops...Sorry but I just checked out the commit history. Seems it is the doc that needs to be updated.

reem:
Thanks!

PR #421, opened 2016-02-07
Fix some grammar

svenstaro:
First "its" is definitely possessive and second one should go due to the first one already expressing possessiveness.


reem:
Thanks!

PR #422, opened 2016-02-09
Generic SSL Interface

jimktrains:
This allows for the SSL Context to be created by the end developer. (Built to allow custom SNI handling.)

### DIFF
--- a/src/iron.rs
+++ b/src/iron.rs
@@ -4,19 +4,17 @@
 use std::net::{ToSocketAddrs, SocketAddr};
 use std::time::Duration;
 #[cfg(feature = "ssl")]
-#[cfg(feature = "openssl")]
 use std::path::PathBuf;
 
 #[cfg(feature = "ssl")]
-#[cfg(feature = "openssl")]
 use hyper::net::{NetworkStream, Openssl, Ssl};
-#[cfg(feature = "ssl")]
-#[cfg(feature = "openssl")]
-use openssl::ssl::{Ssl, SslContext, SslStream, SslMethod, SSL_VERIFY_NONE};
 
 pub use hyper::server::Listening;
 use hyper::server::Server;
-use hyper::net::Fresh;
+use hyper::net::{Fresh, NetworkListener, HttpListener};
+
+#[cfg(feature = "ssl")]
+use hyper::net::HttpsListener;
 
 use request::HttpRequest;
 use response::HttpResponse;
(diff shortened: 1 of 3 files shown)


src/iron.rs
```diff
@@ -6,9 +6,15 @@ use std::time::Duration;
...
 
+#[cfg(feature = "ssl")]
+use hyper::net::{NetworkStream, Openssl, Ssl};
+
 pub use hyper::server::Listening;
 use hyper::server::Server;
-use hyper::net::Fresh;
+use hyper::net::{Fresh, NetworkListener, HttpListener};
+
+#[cfg(feature = "ssl")]
```
sfackler: The ssl feature doesn't seem super necessary.

src/iron.rs
```diff
@@ -87,11 +90,68 @@ impl Protocol {
...
+    fn protocol(&self) -> Protocol { Protocol::Http }
+
+    fn create_server(&self, sock_addr: SocketAddr) -> Result<Server, Error>{
+        Server::http(sock_addr)
+   }
+}
+
+/// Default HTTPs handler
+#[cfg(feature = "ssl")]
+pub struct HttpsProtocolHandler<S: Ssl + Clone + Send> {
```
sfackler: I'd just give this a constructor that takes in the `S` and avoid the Option + `impl HttpsProtocolHandler<Openssl>` altogether.

jimktrains:
Travis is failing because I didn't commit a dependency to openssl in Cargo.toml, but use it in the code. That may need to be refactored out, or the HTTPS server moved somewhere else?

src/iron.rs
jimktrains: This is a hold over from the current codebase; I didn't add the feature.

jimktrains:
Also, I'll eventually squash all of this, or rewrite some of the commit messages. I just realized my first one is non-informative.

src/iron.rs
```diff
@@ -87,11 +90,68 @@ impl Protocol {
...
-            Protocol::Https { .. } => "https"
+            Protocol::Https => "https"
         }
     }
 }
 
+/// Protocol Handlers create servers
+pub trait ProtocolHandler<TListener: NetworkListener> {
+    /// Return the name used for this protocol in a URI's scheme part.
+    fn name(&self) -> &'static str;
```
samgiles: Do we need this if `protocol().name()` returns the same thing?

posix4e:
Do we still want this?

sfackler:
Yes!

reem:
Closing because stale, but please feel free to make a new PR with updated code!

jimktrains:
I was coming back to this, but it looks like there is now a way to pass in a TLS server to handle the TLS-bits.

PR #428, opened 2016-02-14
(feat) Add BodyReader wrapper

panicbit:
This allows to easily set a `Read`er as response body without dealing with `as Box<WriteBody>` shenanigans.

# Example
```rust
let reader = io::empty();
Response::with(BodyReader(reader));
```

I'm not sure if we actually want to implement it this way.
The pros are that we can avoid annoying type wrestling and double-boxing.
The cons are that we need to introduce intermediate types like `BodyReader`.

### DIFF
--- a/src/modifiers.rs
+++ b/src/modifiers.rs
@@ -30,6 +30,7 @@
 //! [rust-modifier](https://github.com/reem/rust-modifier).
 
 use std::fs::File;
+use std::io;
 use std::path::{Path, PathBuf};
 
 use modifier::Modifier;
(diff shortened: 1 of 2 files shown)


reem:
BodyReader needs a public constructor, since its field is private. Otherwise LGTM.

panicbit:
Not sure what you're seeing, but `BodyReader` and its fields look pretty public to me:

```rust
pub struct BodyReader<R: Send>(pub R);
```

;)

reem:
Err, my bad, for some reason I thought the field was private.

PR #429, opened 2016-02-15
(fix) Set status::Ok on expected routes in simple_routing example.

xevix:
The two "hello" routes by default will be set to `status::NotFound`, but the routes return data like a typical request so `status::Ok` seems more appropriate.


untitaker:
Thanks!

PR #432, opened 2016-02-28
Remove ssl as a default feature

### LINKED ISSUE #409: Change SSL to opt-in
Currently the SSL library is opt-out, having to manually specify `default-features=false` in every crate that depends on Iron.

Many of the middleware available does not care about whether SSL is enabled or not. However, many libraries do not opt out of the feature, making compilation without it almost impossible. This is especially an issue on Windows where compiling openssl-sys is not easy.

#351 did not take a clear stand on inclusion of this feature. I would like to see this changed to the opt-in behaviour.

Sample middleware which are currently unsusable without ssl:
* [handlebars-iron](https://github.com/sunng87/handlebars-iron/blob/master/Cargo.toml)
* [router](https://github.com/iron/router) (present in TOML, but not yet pushed to crates.io)
* [mount](https://github.com/iron/mount/blob/master/Cargo.toml)
* [staticfile](https://github.com/iron/staticfile/blob/master/Cargo.toml)

mvdnes:
Resolves #409

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ rust:
     - nightly
 sudo: false
 script:
+    - cargo build
+    - cargo test
     - cargo build --features=ssl
     - cargo test --features=ssl
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ rust:
     - nightly
 sudo: false
 script:
-    - cargo build
-    - cargo test
+    - cargo build --features=ssl
+    - cargo test --features=ssl
 env:
     global:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,8 +4,8 @@ rust:
     - nightly
 sudo: false
 script:
-    - cargo build
-    - cargo test
+    - cargo build --features=ssl
+    - cargo test --features=ssl
```
reem: Can you add `cargo build` and `cargo test` without `ssl` as well?

mvdnes:
Pushed a commit for testing with and without the ssl feature.

untitaker:
@reem? LGTM

mvdnes:
Great, thanks! :+1:

PR #433, opened 2016-03-01
Allow the Servers to be created in a generic way

samgiles:
This is really similar to @jimktrains's  PR - but with the suggestion addressed.

This is to support server side SNI with Iron - which I've implemented over here https://github.com/samgiles/foxbox/blob/https-support/src/http_server.rs#L75.  I plan on moving that into a nicer, and reusable module if this gets merged.

### DIFF
--- a/src/iron.rs
+++ b/src/iron.rs
@@ -8,7 +8,13 @@ use std::path::PathBuf;
 
 pub use hyper::server::Listening;
 use hyper::server::Server;
-use hyper::net::Fresh;
+use hyper::net::{Fresh, NetworkListener, HttpListener};
+
+
+#[cfg(feature = "ssl")]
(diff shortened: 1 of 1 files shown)


samgiles:
Hey @reem, It would be awesome to get this feature merged, I'd be happy to make changes if you see any.

untitaker:
I feel like an API like this would be less Java-esque:

    let iron = Iron::new(h);
    iron.protocol(Http);
    iron.addr(...);
    iron.hyper_server(...)  // The actual new functionality
    iron.listen();

Since the existing methods `http`, `https` and `listen_with` all consume `self`, they can modify the `Iron` instance at will.

samgiles:
Hmm, nice idea, will take a look at doing that tomorrow, thanks.

reem:
I agree with @untitaker, closing for now but feel free to make a new PR any time with the new plan.

PR #434, opened 2016-03-10
Expand docs on modifiers

dbrgn:
I hope this should make them more accessible to newbies (like me).

### DIFF
--- a/src/modifiers.rs
+++ b/src/modifiers.rs
@@ -1,30 +1,47 @@
-//! Defines a series of convenience modifiers for editing Responses
+//! This module defines a series of convenience modifiers for changing
+//! Responses.
 //!
-//! Modifiers can be used to edit Responses through the owning
-//! method `set` or the mutating `set_mut`, both of which are
-//! defined in the `Set` trait.
+//! Modifiers can be used to edit `Response`s through the owning method `set`
+//! or the mutating `set_mut`, both of which are defined in the `Set` trait.
 //!
-//! Instead of having a combinatorial explosion of Response methods
-//! and constructors, this provides a series of modifiers that
-//! can be used through the `Set` trait.
+//! For Iron, the `Modifier` interface offers extensible and ergonomic response
+//! creation while avoiding the introduction of many highly specific `Response`
+//! constructors.
 //!
-//! For instance, instead of `Response::redirect` constructing a
-//! redirect response, we provide a `Redirect` modifier, so you
-//! can just do:
+//! The simplest case of a modifier is probably the one used to change the
+//! return status code:
 //!
 //! ```
 //! # use iron::prelude::*;
 //! # use iron::status;
-//! # use iron::modifiers::Redirect;
-//! # use iron::Url;
-//!
-//! let url = Url::parse("http://doc.rust-lang.org").unwrap();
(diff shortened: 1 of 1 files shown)


src/modifiers.rs
```diff
@@ -1,4 +1,5 @@
-//! Defines a series of convenience modifiers for editing Responses
+//! This module defines a series of convenience modifiers for editing
```
steveklabnik: I might use 'change' rather than 'edit' here, for consistency with the rest of everything.

src/modifiers.rs
```diff
@@ -8,24 +9,48 @@
...
 //! # use iron::prelude::*;
 //! # use iron::status;
-//! # use iron::modifiers::Redirect;
+//!
+//! let r = Response::with(status::NotFound);
+//! assert_eq!(r.status.unwrap().to_u16(), 404);
+//! ```
+//!
+//! You can also pass in a tuple of modifiers, they will all be applied. Here's
+//! an example of a modifier 2-tuple that will change the status code and the
```
steveklabnik: tutuple, hehe
tutu

steveklabnik:
This is very helpful :+1:

src/modifiers.rs
```diff
@@ -8,24 +9,45 @@
...
+//! # use iron::prelude::*;
+//! # use iron::status;
+//! # use iron::modifiers;
+//! # use iron::Url;
 //! let url = Url::parse("http://doc.rust-lang.org").unwrap();
-//! Response::with((status::Found, Redirect(url)));
+//! Response::with((status::Found, modifiers::Redirect(url)));
 //! ```
 //!
 //! This is more extensible as it allows you to combine
```
SkylerLipthay: Should this paragraph be moved, perhaps to after line 10?

src/modifiers.rs
dbrgn: Hahaha :)

Do you think just `tuple` would be better?

src/modifiers.rs
```diff
@@ -1,30 +1,49 @@
...
+//! or the mutating `set_mut`, both of which are defined in the `Set` trait.
 //!
-//! Instead of having a combinatorial explosion of Response methods
-//! and constructors, this provides a series of modifiers that
-//! can be used through the `Set` trait.
+//! Instead of having a combinatorial explosion of Response methods and
+//! constructors, this provides a series of modifiers that can be used through
+//! the `Set` trait. This is more extensible as it allows you to combine
+//! arbitrary modifiers without having a massive number of Response
+//! constructors.
```
dbrgn: @SkylerLipthay I moved the paragraph up here, but still think that the wording is a bit confusing. Maybe you or @steveklabnik have a better proposal on wording?

SkylerLipthay: I think the whole paragraph could be replaced with "For Iron, the `Modifier` interface offers extensible and ergonomic `Response` creation while avoiding the introduction of many highly specific `Response` constructors."

dbrgn: Sounds great!

dbrgn:
Fixed and rebased.

reem:
Can we add back a note somewhere that you can write your own modifiers? This is a big reason for their use.

PR #436, opened 2016-03-15
Use 'static instead of Any for Handler since we don't need Reflect

fabricedesre:
That's removing the constraint of Reflect on Handler implementations, which is sometimes useful.


untitaker:
Thanks!

PR #439, opened 2016-03-16
Update hyper to 0.8

AerialX:
One particular annoyance is that hyper 0.7 pulls in a version of mime that unnecessarily pulls in serde as a dependency.


untitaker:
Thanks!

PR #445, opened 2016-03-29
Url Display Refactor

### LINKED ISSUE #442: Hide default port in `Display` trait for `Url`.
Because the `Display` is meant to be user facing, I'd like it if default ports (80 for `http` and 443 for `https`) are hidden. Currently the following is true:

```rust
assert_eq!(format!("{}", Url::parse("http://example.com").unwrap()), "http://example.com:80".to_owned());
assert_eq!(format!("{}", Url::parse("https://example.com").unwrap()), "https://example.com:443".to_owned());
```

Instead could it be the following?

```rust
assert_eq!(format!("{}", Url::parse("http://example.com").unwrap()), "http://example.com".to_owned());
assert_eq!(format!("{}", Url::parse("https://example.com").unwrap()), "https://example.com".to_owned());
```

gsquire:
Fixes #442 
I also added a few tests to ensure the conditional is covered correctly.

### DIFF
--- a/src/request/url.rs
+++ b/src/request/url.rs
@@ -152,9 +152,12 @@ impl fmt::Display for Url {
         // Write the host.
         try!(self.host.fmt(formatter));
 
-        // Write the port.
-        try!(":".fmt(formatter));
-        try!(self.port.fmt(formatter));
+        // Write the port if they are not the default ports of 80 for HTTP and 443 for HTTPS.
+        if !((self.scheme == "http".to_owned() && self.port == 80) ||
+            (self.scheme == "https".to_owned() && self.port == 443)) {
+                try!(":".fmt(formatter));
+                try!(self.port.fmt(formatter));
+        }
 
         // Write the path.
         try!(write!(formatter, "{}", PathFormatter { path: &self.path }));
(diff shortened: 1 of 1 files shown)


src/request/url.rs
```diff
@@ -152,9 +152,12 @@ impl fmt::Display for Url {
         // Write the host.
         try!(self.host.fmt(formatter));
 
-        // Write the port.
-        try!(":".fmt(formatter));
-        try!(self.port.fmt(formatter));
+        // Write the port if they are not the default ports of 80 for HTTP and 443 for HTTPS.
+        if !((self.scheme == "http".to_owned() && self.port == 80) ||
+            (self.scheme == "https".to_owned() && self.port == 443)) {
```
fsommar: Wouldn't it be possible to do &self.scheme == "http" or similarly to prevent allocations?

gsquire: You're right, plus that looks cleaner. Good catch.

gsquire:
@untitaker If this looks good, I can rebase my commits.

untitaker:
Thanks!

Issue #261, opened 2015-01-18
Split hyper from iron core


sunng87:
Currently, our http server hyper is built into core parts of Iron. If you look at Python's WSGI or Clojure's Ring, there is a protocol between webapp and web server. And the server implementation are made separated from the protocol and webapp. In Clojure, for example, the protocol is [ring](https://github.com/ring-clojure/ring), it's middleware based like Iron. It comes with a default server implementation ring-jetty-adapter, but it's totally decoupled from ring, so you can write servers and make adapters for ring. In python, there's also similar framework like bottle.

Although hyper is the only http server currently, I think it's still necessary to split and make each part easy to test.

reem:
I agree completely. I haven't really felt the need to do this yet since everything is still in flux and hyper is the only real option right now, but eventually I think there should be a protocol/trait layer between hyper and Iron so that you can swap out the hyper back end.

PR #450, opened 2016-04-13
(refactor) Add a new public listening function for custom Server

### LINKED ISSUE #447: Advanced OpenSSL config (client certs)
I want to use Iron in a context where I'm required to use client certificates.

However, the OpenSSL config is quite limited now, and I'm not able to set any of the advanced configuration options.

I'd like to contribute, but I'm not sure what the best approach would be. I'm considering adding a struct to `Protocol::Https`, but that would break backwards compatibility with existing code.

I can also add a new enum variant, something like `Protocol::AdvancedHttps` with an additional struct with advanced OpenSSL options (such as client certs).

What do you think?

rix0rrr:
This allows consumers to set up the server as desired, such as configuring the
SSL context for client certificates.

The API is a little iffy, since consumers still need to pass in an instance of
Protocol with certificate and key, which will be unused.

Fixes #447, and forms a starting point for further abstraction in #261.

### DIFF
--- a/src/iron.rs
+++ b/src/iron.rs
@@ -140,23 +140,17 @@ impl<H: Handler> Iron<H> {
     ///
     /// Panics if the provided address does not parse. To avoid this
     /// call `to_socket_addrs` yourself and pass a parsed `SocketAddr`.
-    pub fn listen_with<A: ToSocketAddrs>(mut self, addr: A, threads: usize,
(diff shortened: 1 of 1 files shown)


untitaker:
If we merge this, I wonder if we would have to nuke this API again when we refactor the entire thing for Hyper 0.10 (async API).

sfackler:

untitaker:
:+1:

PR #451, opened 2016-04-26
Don't deny any lints

bluss:
Don't deny any lints

This makes iron forward-compatible with lint changes.

(We can't use deny(missing_docs) for test builds due to bug rust-lang/rust/issues/24584.)

This is for the same reason as rust-lang/cargo/pull/2622. It is of course best to wait for their lead and see what other rust & cargo devs think.

Iron is now privileged to be part of rust's integration tests. As a counter gesture, please don't have iron break on lint changes and improvements.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,6 @@
 #![doc(html_logo_url = "https://avatars0.githubusercontent.com/u/7853871?s=128", html_favicon_url = "https://avatars0.githubusercontent.com/u/7853871?s=256", html_root_url = "http://ironframework.io/core/iron")]
 #![cfg_attr(test, deny(warnings))]
-#![deny(missing_docs)]
+#![warn(missing_docs)]
 
 //! The main crate for Iron.
 //!


untitaker:
Contrary I don't think that it should be possible to release a new Iron version where such lints fail (improved or not).

I see that the Cargo issue got closed.

bluss:
Yeah I don't completely understand how they decided to solve it. But it looks like upstream rust now ignores lint-related failures in cargotest, which means that iron could stop building due to lint changes upstream, and cargotest does not block on that.

bluss:
I will underline that such lint changes are most often improvements, as in for example, finding more unused items than before.

reem:
@bluss, could you change this to instead still deny missing docs on test? My intention is to force travis to fail if anything is undocumented.

bluss:
@reem thanks for looking at it. I think I want to abandon this now. iron is a library and cargo is a binary, so it's a bit different anyway.

PR #452, opened 2016-05-02
Update url to 1.1 and hyper to 0.9

carols10cents:
Hi! ❤️ So the url crate now has a 1.0 version out that has some breaking changes, and hyper 0.9 is out, which uses url 1.0, and I think because hyper does `pub use url::Url`, iron can't update hyper without updating url. So I decided to give it a try!

I put most of the details of the relevant changes in [the commit message for the last commit](https://github.com/iron/iron/commit/bca05d75107e7bb3896064e78633334d663d5045). There might be better ways to do the conversions between Iron's Url and url's Url, but because the url crate now does a lot of checks and coordination in the setters, I didn't think it would be a good idea to set the fields of the Url directly.

If @SimonSapin has time, he might want to code review this and make sure I'm not missing anything :)

### DIFF
--- a/src/request/url.rs
+++ b/src/request/url.rs
@@ -67,42 +67,41 @@ impl Url {
     /// Create a `Url` from a `rust-url` `Url`.
     pub fn from_generic_url(raw_url: url::Url) -> Result<Url, String> {
         // Create an Iron URL by extracting the special scheme data.
-        match raw_url.cannot_be_a_base() {
-            false => {
-                // Extract the port as a 16-bit unsigned integer.
-                let port: u16 = match raw_url.port_or_known_default() {
-                    // If explicitly defined or has a known default, unwrap it.
-                    Some(port) => port,
+        if raw_url.cannot_be_a_base() {
+            Err(format!("Not a special scheme: `{}`", raw_url.scheme()))
+        } else {
+            // Extract the port as a 16-bit unsigned integer.
+            let port: u16 = match raw_url.port_or_known_default() {
+                // If explicitly defined or has a known default, unwrap it.
+                Some(port) => port,
 
-                    // Otherwise, use the scheme's default port.
-                    None => return Err(format!("Invalid special scheme: `{}`",
-                                                    raw_url.scheme())),
-                };
-                // Map empty usernames to None.
-                let username = match raw_url.username() {
-                    "" => None,
-                    _ => Some(raw_url.username())
-                };
-                // Map empty passwords to None.
-                let password = match raw_url.password() {
-                    None => None,
-                    Some(ref x) if x.is_empty() => None,
-                    Some(password) => Some(password)
-                };
-                Ok(Url {
-                    scheme: raw_url.scheme().to_string(),
-                    // `unwrap` is safe here because urls that cannot be a base don't have a host
-                    host: raw_url.host().unwrap().to_owned(),
-                    port: port,
-                    // `unwrap` is safe here because urls that can be a base will have `Some`.
-                    path: raw_url.path_segments().unwrap().map(str::to_string).collect(),
-                    username: username.map(str::to_string),
-                    password: password.map(str::to_string),
-                    query: raw_url.query().map(str::to_string),
-                    fragment: raw_url.fragment().map(str::to_string),
-                })
-            },
-            true => Err(format!("Not a special scheme: `{}`", raw_url.scheme()))
+                // Otherwise, use the scheme's default port.
+                None => return Err(format!("Invalid special scheme: `{}`",
+                                                raw_url.scheme())),
+            };
+            // Map empty usernames to None.
+            let username = match raw_url.username() {
+                "" => None,
+                _ => Some(raw_url.username())
+            };
+            // Map empty passwords to None.
+            let password = match raw_url.password() {
+                None => None,
+                Some(ref x) if x.is_empty() => None,
+                Some(password) => Some(password)
+            };
+            Ok(Url {
+                scheme: raw_url.scheme().to_string(),
+                // `unwrap` is safe here because urls that cannot be a base don't have a host
+                host: raw_url.host().unwrap().to_owned(),
+                port: port,
+                // `unwrap` is safe here because urls that can be a base will have `Some`.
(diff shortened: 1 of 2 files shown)


src/request/url.rs
```diff
@@ -70,70 +67,47 @@ impl Url {
...
+                    scheme: raw_url.scheme().to_string(),
+                    // `unwrap` is safe here because urls that cannot be a base don't have a host
+                    host: raw_url.host().unwrap().to_owned(),
                     port: port,
-                    path: data.path,
-                    username: username,
-                    password: password,
-                    query: raw_url.query,
-                    fragment: raw_url.fragment
+                    path: raw_url.path_segments().into_iter().flat_map(|x| x).map(|x| x.to_string()).collect(),
```
SimonSapin: Given `raw_url.cannot_be_a_base() == false`, `raw_url.path_segments()` can be assumed to return `Some`. So there is no need to use `Option::into_iter` and this line can be `path: raw_url.path_segments().unwrap().map(|x| x.to_string()).collect()`.

src/request/url.rs
```diff
@@ -70,70 +67,47 @@ impl Url {
...
+                    username: username.map(|s| s.to_string()),
+                    password: password.map(|s| s.to_string()),
+                    query: raw_url.query().map(|s| s.to_string()),
+                    fragment: raw_url.fragment().map(|s| s.to_string()),
```
LFalch: Could this not just be `std::string::ToString::to_string`?

untitaker:
IIRC @reem mentioned that he wanted to refactor the API around Iron requests/urls as well, I hope this will eliminate all those costly conversions.

src/request/url.rs
carols10cents: Sorry, do you mean make this change?

```diff
- fragment: raw_url.fragment().map(|s| s.to_string()),
+ fragment: raw_url.fragment().to_string(),
```

This doesn't work because `url::Url::fragment` returns `Option<&str>`, and `Option<&str>` doesn't implement `Display`. Or is there another way to use `std::string::ToString::to_string`?

untitaker: You can pass ToString::to_string to map.

On 3 May 2016 21:44:39 CEST, "Carol (Nichols || Goulding)" <notifications@github.com> wrote:
>>                      port: port,
>> -                    path: data.path,
>> -                    username: username,
>> -                    password: password,
>> -                    query: raw_url.query,
>> -                    fragment: raw_url.fragment
>> +                    path:
>raw_url.path_segments().into_iter().flat_map(|x| x).map(|x|
>x.to_string()).collect(),
>> +                    username: username.map(|s| s.to_string()),
>> +                    password: password.map(|s| s.to_string()),
>> +                    query: raw_url.query().map(|s| s.to_string()),
>> +                    fragment: raw_url.fragment().map(|s|
>s.to_string()),
>
>Sorry, do you mean make this change?
>
>```diff
>- fragment: raw_url.fragment().map(|s| s.to_string()),
>+ fragment: raw_url.fragment().to_string(),
>```
>
>This doesn't work because `url::Url::fragment` returns `Option<&str>`,
>and `Option<&str>` doesn't implement `Display`. Or is there another way
>to use `std::string::ToString::to_string`?
>
>---
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/iron/iron/pull/452/files/bca05d75107e7bb3896064e78633334d663d5045#r61943140

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

src/request/url.rs
carols10cents: Good catch! I've added a commit to address this. Thank you!

carols10cents:
> IIRC @reem mentioned that he wanted to refactor the API around Iron requests/urls as well, I hope this will eliminate all those costly conversions.

Yes, I think there's definitely room for improvement. Perhaps iron could hold onto the `url::Url` within its `Url` type instead of converting?

src/request/url.rs
```diff
@@ -70,70 +67,48 @@ impl Url {
     /// Create a `Url` from a `rust-url` `Url`.
     pub fn from_generic_url(raw_url: url::Url) -> Result<Url, String> {
         // Create an Iron URL by extracting the special scheme data.
-        match raw_url.scheme_data {
-            SchemeData::Relative(data) => {
+        match raw_url.cannot_be_a_base() {
```
reem: match on bool should be an if/else

reem:
I need to catch up on upstream `url` changes to know if we could just hold on to `url::Url` directly, but if it is possible then I would love to do it.

src/request/url.rs
untitaker: Though I am unsure if that's a desirable change or purely subject to
bikeshedding.

SimonSapin: Maybe @LFalch meant replacing `.map(|s| s.to_string())` with `.map(::std::string::ToString::to_string)`? In that case it can also be written `.map(str::to_string)` (since the `str` type is in the prelude), but I don’t know if it’s better than a closure. Matter of preference, I suppose.

SimonSapin:
@reem let me know if/how changing rust-url would help. (Though preferably non-breaking changes at this point.)

src/request/url.rs
carols10cents: Ahhhhh I see! I'll add a commit on here and yinz can choose to keep it or not :)

LFalch: @SimonSapin That's exactly what I meant!

src/request/url.rs
carols10cents: On it!

carols10cents: And commit added.

carols10cents:
So is merging this in dependent on trying the refactoring to hold on to `url::Url` directly? If so, mind if I give it a try?

carols10cents:
Forgiveness > permission... https://github.com/iron/iron/pull/453

s-panferov:
Any progress on this? I'm very interested in this PR to update [rustless](https://github.com/rustless/rustless) on newer `hyper` too. Can I help you in something?

reem:
Lets move all further discussion to #453, which is a superset of this PR (and hopefully can land soon!)

PR #453, opened 2016-05-06
Hold onto rust-url Url in iron Url

carols10cents:
Rather than doing so many conversions back and forth.

This builds on #452 and contains 3 commits more than that PR.

Commits eea2bbf and bef95ae are fairly straightforward. eea2bbf gets rid of the extra `rust-url` parse call needed since `rust-url` 1.0 to get back to a `rust-url` `Url`. bef95ae uses `rust-url`'s display implementation on the stored `rust-url` `Url` for the `iron` `Url`'s display trait-- all the tests still pass.

3a9b322 may be a step too far-- I'm happy to remove it if so. It changes the public API of `iron` `Url` to have methods instead of the struct fields, and those methods mostly just delegate to the `rust-url` `Url`. I'm not sure if this is what you had in mind or not!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ ssl = ["hyper/ssl"]
 
 [dependencies]
 typemap = "0.3"
-url = "1.0"
+url = "1.1"
 plugin = "0.2"
 modifier = "0.1"
 error = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ ssl = ["hyper/ssl"]
 
 [dependencies]
 typemap = "0.3"
-url = "0.5"
+url = "1.0"
 plugin = "0.2"
 modifier = "0.1"
 error = "0.1"
@@ -35,7 +35,7 @@ lazy_static = "0.1"
 num_cpus = "0.2"
 
 [dependencies.hyper]
-version = "0.8"
+version = "0.9"
 default-features = false
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


reem:
Just gave this a more thorough read, and it looks fantastic! Thank you very much for making this change, it's something I've wanted to do for a long time.

There's only one major thing I think we should change: many of the methods return owned types (`String`, `Vec<String>`, etc.) by internally calling cloning methods when we could just return the borrowed variant. Wherever possible, I would like to avoid adding any allocations by just returning the borrowed type instead.

I think the only remaining refactor would be substituting String for a proper error type, but that can be left for another PR.

carols10cents:
@reem awesome, done! That makes the implementations way nicer-- fewer `.to_owned()`s all over the place :)

onur:
Any progress on this? I am getting this warning when I try to build iron with latest cargo (5a26b65):

```
$ cargo build
warning: TOML file found which contains invalid syntax and will soon not parse
at `/home/onur/.cargo/registry/src/github.com-1ecc6299db9ec823/url-0.5.9/Cargo.toml`.

The TOML spec requires newlines after table definitions (e.g. `[a] b = 1` is
invalid), but this file has a table header which does not have a newline after
it. A newline needs to be added and this warning will soon become a hard error
in the future.
```

And I can't already build iron 0.3 with cratesfyi. We really need url 1 support soon I guess.

bfrog:
I'm going to spend sometime this weekend working on trying to update iron to work with hyper master, I'm basing my changes on this PR now because I believe this is just waiting to merged @reem

carols10cents:
YAAAYY!!!! 🎆🎆🎆🎆🎆🎆 Thank you @reem !!! ❤️ ❤️ ❤️ ❤️

PR #455, opened 2016-05-12
make RedirectRaw's field public to allow construction

cmsd2:
RedirectRaw is necessary to allow redirection to urls which don't have special schemes (and therefore are incompatible with Redirect) such as oob:// and steam://

This patch allows construction of RedirectRaw by making the inner field public.


reem:
thanks!

PR #457, opened 2016-05-23
Add methods to launch an http or https server on a bound Listener

zofrex:
Another attempt at #386. Please also see #415 which addresses the same issue.

I tried to keep things so no hyper internals are exposed, and I refactored a little to try to reduce the repetition that otherwise snuck in.

Example usage:

``` rust
extern crate iron;

use iron::prelude::*;
use iron::status;
use std::net::TcpListener;

fn main() {
    fn hello_world(_: &mut Request) -> IronResult<Response> {
        Ok(Response::with((status::Ok, String::from("Hello World!"))))
    }

    let listener = TcpListener::bind("0.0.0.0:80").unwrap();

    let _i = Iron::new(hello_world).http_on(listener).unwrap();
}
```

### DIFF
--- a/src/iron.rs
+++ b/src/iron.rs
@@ -8,7 +8,9 @@ use std::path::PathBuf;
 
 pub use hyper::server::Listening;
 use hyper::server::Server;
-use hyper::net::Fresh;
+use hyper::net::{NetworkListener, HttpListener, Fresh, NetworkStream};
+#[cfg(feature = "ssl")]
+use hyper::net::{HttpsListener};
 
 use request::HttpRequest;
 use response::HttpResponse;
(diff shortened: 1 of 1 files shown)


Hoverbear:
@untitaker I remember you saying something about a PR like this (I can't actually find the comment) might need it's interface nuked in the upcoming async branch? Do you have thoughts on merging something like this or 415?

untitaker:
I was thinking more about how creating and returning responses (especially streaming responses) would work within handlers. `.listen`, `.https` etc may not differ that much, but I'd still wait for Hyper's async stuff to stabilize.

I think you mean #450 in particular, I just don't know how Hyper's API will develop.

untitaker:
You could argue though that we shouldn't let development stall while Hyper's figuring out how to do async.

Hoverbear:
Assigning myself to review.

sfackler:

untitaker:
Great!

PR #475, opened 2016-08-19
Improve conversion of iron::Url from/to url::Url

untitaker:
When I want to read data from the underlying `url::Url` (either because
it offers a richer API than Iron's URL for my usecases, or because I can
pass it directly into another API), I currently have two choices:

- `let url = request.url.into_generic_url().clone()`. This unnecessarily
  copies data.

- `let url = request.url.into_generic_url()`, which moves `request.url`,
  and do `request.url = iron::Url::from_generic_url(url)` when I'm done,
  to move the generic URL back into the request. This is unnecessarily
  verbose.

Both options are undesirable. Therefore `generic_url`.

I also have similar usecase for `generic_url_mut`. I have a middleware
that modifies a part of the request URL. Currently I'd have to do
similar tricks to do that.

### DIFF
--- a/src/request/url.rs
+++ b/src/request/url.rs
@@ -42,18 +42,9 @@ impl Url {
     }
 
     /// Create a `rust-url` `Url` from a `Url`.
+    #[deprecated(since="0.5.0", note="use `into` from the `Into` trait instead")]
     pub fn into_generic_url(self) -> url::Url {
         self.generic_url
-    }
-
-    /// Return an immutable pointer to the underlying `rust-url` `Url`.
-    pub fn generic_url(&self) -> &url::Url {
-        &self.generic_url
-    }
-
-    /// Return a mutable pointer to the underlying `rust-url` `Url`.
-    pub fn generic_url_mut(&mut self) -> &mut url::Url {
-        &mut self.generic_url
     }
 
     /// The lower-cased scheme of the URL, typically "http" or "https".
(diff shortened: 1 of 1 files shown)


Hoverbear:
I guess it wouldn't make sense to use `Into` for this since you just want a pointer.

untitaker:
I can imagine using Into instead of `into_generic_url` though. AsRef would probably be useful.

Hoverbear:
Definitely. :)

untitaker:
Would you say that implementing `From<url::Url>`, `Into<url::Url>` and `AsRef<url::Url>` is preferrable? Should I remove the redundant methods or leave those that are already in a stable release?

Hoverbear:
If it already exists we should keep it -- People may depend on it.

But using existing, idiomatic methods such as `Into` etc would definitely be preferable for new code.

I, for one, prefer simple, familiar concepts over library-specific concepts. If you prefer this method it'd be awesome if we could have both ways.

untitaker:
>But using existing, idiomatic methods such as Into etc would definitely be preferable for new functions.

I agree.

>it'd be awesome if we could have both ways.

I disagree, I don't think we should duplicate the API, and deprecate the existing methods.

I pushed a commit to show what I mean.

untitaker:
We can't actually implement `From<url::Url>` because `from_generic_url` returns a `Result` and may fail. `TryFrom` is unfortunately not stable.

Hoverbear:
@untitaker It looks very nice regardless. :)

untitaker:
Do we agree though that `into_generic_url` should be deprecated? @reem Any comments?

untitaker:
@reem I think this is ready to merge.

untitaker:
I'll just go ahead...

PR #485, opened 2016-09-30
(example) Add HTTPS example.

Hoverbear:
Needs to use cfg flags because of the 'ssl' feature not being default.
If we don't do this then `cargo test` will fail.

@untitaker can you review and merge if it's good?

### DIFF
--- a/examples/https.rs
+++ b/examples/https.rs
@@ -30,7 +30,7 @@ fn main() {
         Ok(Response::with((status::Ok, "Hello world!")))
     }).https("127.0.0.1:3000", cert, key) {
         Result::Ok(listening) => println!("{:?}", listening),
-        Result::Err(err) => println!("{:?}", err),
+        Result::Err(err) => panic!("{:?}", err),
     }
     // curl -vvvv https://127.0.0.1:3000/ -k
 }
--- a/examples/https.rs
+++ b/examples/https.rs
@@ -1,0 +1,41 @@
+// This requires running with:
+//
+// ```bash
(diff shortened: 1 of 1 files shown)


Hoverbear:
`ci/circleci` fails with a 404 and is not actually a test failure.

Hoverbear:
Travis failures are due to temporary rustup breakage. (Ref: https://github.com/rust-lang/rustup.sh/pull/73)

examples/https.rs
```diff
@@ -0,0 +1,41 @@
...
+    // openssl genrsa -out localhost.key 4096
+    let key = Path::new("localhost.key").to_path_buf();
+    // openssl req -key localhost.key -x509 -new -days 3650 -out localhost.crt
+    let cert = Path::new("localhost.crt").to_path_buf();
+
+    match Iron::new(|_: &mut Request| {
+        Ok(Response::with((status::Ok, "Hello world!")))
+    }).https("127.0.0.1:3000", cert, key) {
+        Result::Ok(listening) => println!("{:?}", listening),
+        Result::Err(err) => println!("{:?}", err),
```
untitaker: Needs a different exitcode I think. Also for illustratory purposes I think it's good enough to `unwrap` to keep the example short.

Hoverbear: This method is nice because it outputs `Listening { socket: V4(127.0.0.1:3000) }` which is nice for users to see "Oh hey look it's listening I can browse to it".

untitaker:
Cool stuff, thanks.

PR #490, opened 2016-10-27
Don't mix up local address and Host header

lfairy:
Most of the time, when the Host header and local address don't match, it's because we're behind a reverse proxy. In this case, using the local port (which is likely an internal port such as "8080") doesn't make sense.


untitaker:
Lgtm

PR #491, opened 2016-10-27
(doc) Create complex_routing.rs

bspeice:
Add a slightly more nuanced and realistic routing example.
I'm not sure if this should be committed to the router repository, as it technically depends on code from there.
Intended to address issue #311

### DIFF
--- a/examples/complex_routing.rs
+++ b/examples/complex_routing.rs
@@ -1,0 +1,31 @@
+extern crate iron;
+extern crate router;
+
+use iron::Handler;
+use iron::status;
+use iron::IronResult;
+use iron::Response;
+use iron::Request;
+use iron::Iron;
+use router::Router;
+
+struct EchoHandler {
+    message: String
+}
+
+impl Handler for EchoHandler {
+    fn handle(&self, _: &mut Request) -> IronResult<Response> {
+        Ok(Response::with((status::Ok, self.message.clone())))
+    }
+}
+
+fn main() {
+    let echo = EchoHandler {
+        message: "You've found the index page!".to_string()
+    };
+
+    let mut router = Router::new();
+    router.get("/", echo, "index");
+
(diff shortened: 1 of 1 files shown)


untitaker:
Thanks for this PR, however I don't understand, why is this more realistic?

bspeice:
I couldn't find any examples on how to route to things that weren't static functions. This adds the ability to encapsulate other state that can be used in responses

examples/complex_routing.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use iron::Handler;
+use iron::status;
+use iron::IronResult;
+use iron::Response;
+use iron::Request;
+use iron::Iron;
+use router::Router;
+
+struct EchoHandler {
```
untitaker: This doesn't "echo" anything the client sent.

untitaker:
Ah, the import of routing fails, I think this PR should be filed against the routing repo since the crate is available there.

bspeice:
OK. It shouldn't be hard to do the same example in both cases, I'll get the pull requests whipped up later today.

bspeice:
Closing, as there's already an example of struct-based responses in the routing example.

untitaker:
You mean your PR?

bspeice:
No, the actual [examples/simple_routing.rs](https://github.com/iron/iron/blob/master/examples/simple_routing.rs) uses a struct-based handler. I don't know if this would qualify as redundant.
It's pretty trivial to remove the routing dependency, I can go ahead and clean it up here if you want.

drusellers:
Although this has been closed, this is a huge need. Iron has great support for static functions, but the next step is how do I get my, for lack of a better word, `controller` vibe going on?

untitaker:
There's an example in the routing repo.

PR #492, opened 2016-10-27
Correct requests/sec avg in README

mschmo:
According to the output here: https://github.com/iron/iron/wiki/How-to-Benchmark-hello.rs-Example

> Requests/sec:  72479.41


untitaker:
Fair enough, thanks.

PR #496, opened 2016-11-03
(feat) Added prepare_chain to prepare Chain for middleware

pocket7878:
Sometimes want to write abstract middleware which depends on other middleware on chain.
So, add function to prepare chain before insert middleware it-self to chain.


untitaker:
I think it was intended to lkeep the middleware system independent of Chain, but I don't think there's much value to doing that anyway.

However, if you want to arbitrarily modify the chain, I rather recommend passing the chain as mutable reference *to* your library, instead of passing your middleware to the chain. Instead of exporting a middleware, you'd export a function or method that takes a `&mut Chain` and modifies it.

pocket7878:
@untitaker Thank you for your comment. Ok, I'll take that approach!

PR #510, opened 2017-01-08
Refactor server startup APIs and Https support

### LINKED ISSUE #424: Please help: SNI plus client cert SN
I'm having a little trouble reverse engineering iron and hyper to figure out how to:
- do SNI
- get the verified client cert SN (CN) for authentication

Any pointers would be greatly appreciated.

Thanks!

### LINKED ISSUE #447: Advanced OpenSSL config (client certs)

### LINKED ISSUE #483: ssl feature is not documented in online docs
Struggling to figure out how to start up Iron with HTTPS rather than the default HTTP. Do we also need to provide security certificates, or is that generated by Hyper?

sfackler:
The previous support for SSL was not great - you were locked into
OpenSSL and had no ability to configure it the way you wanted. Hyper is
going to be cutting a release soon that completely removes the built-in
OpenSSL and Secure Transport functionalty in favor of external crates.

While we're at it, also allow a `NetworkListener` to be manually
created, and refactor the configuration a bit to avoid combinatoric
blowup in methods.

Closes #483
Closes #447
Closes #424

r? @untitaker 

It probably makes sense to land this along with the hyper upgrade. @seanmonstar, do you have an idea on when that'll be roughly?

The optional OpenSSL dependency hopefully won't cause problems, but it seemed important to keep the https example around.

### DIFF
--- a/examples/https.rs
+++ b/examples/https.rs
@@ -1,41 +1,41 @@
 // This requires running with:
 //
 // ```bash
-// cargo run --example https --features ssl
+// cargo run --example https --features native-tls-example
 // ```
 //
-// Generate a key and certificate like so:
+// Generate an identity like so:
 //
 // ```bash
-// openssl genrsa -out localhost.key 4096
-// openssl req -key localhost.key -x509 -new -days 3650 -out localhost.crt
+// openssl req -x509 -newkey rsa:4096 -nodes -keyout localhost.key -out localhost.crt -days 3650
+// openssl pkcs12 -export -out identity.p12 -inkey localhost.key -in localhost.crt --password mypass
+//
 // ```
 
 extern crate iron;
+#[cfg(feature = "native-tls-example")]
+extern crate hyper_native_tls;
 
-#[cfg(feature = "ssl")]
+#[cfg(feature = "native-tls-example")]
 fn main() {
-    // Avoid unused errors due to conditional compilation ('ssl' feature is not default)
-    use iron::status;
+    // Avoid unused errors due to conditional compilation ('native-tls-example' feature is not default)
+    use hyper_native_tls::NativeTlsServer;
     use iron::{Iron, Request, Response};
-    use std::path::{Path};
-    use std::result::{Result};
+    use iron::status;
+    use std::result::Result;
 
-    // openssl genrsa -out localhost.key 4096
-    let key = Path::new("localhost.key").to_path_buf();
-    // openssl req -key localhost.key -x509 -new -days 3650 -out localhost.crt
-    let cert = Path::new("localhost.crt").to_path_buf();
+    let ssl = NativeTlsServer::new("identity.p12", "mypass").unwrap();
 
     match Iron::new(|_: &mut Request| {
         Ok(Response::with((status::Ok, "Hello world!")))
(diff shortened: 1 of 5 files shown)


sfackler:
Oops, wrong @seanmonstar..

examples/https.rs
```diff
@@ -1,41 +1,63 @@
...
+#[cfg(feature = "openssl-example")]
 fn main() {
-    // Avoid unused errors due to conditional compilation ('ssl' feature is not default)
-    use iron::status;
+    // Avoid unused errors due to conditional compilation ('openssl-example' feature is not default)
+    use hyper_openssl::OpensslServer;
     use iron::{Iron, Request, Response};
-    use std::path::{Path};
-    use std::result::{Result};
+    use iron::status;
```
untitaker: This new usage is super-verbose. Before we make people change their code, could we have a shortcut for this in e.g. `hyper-openssl`? Something like `OpensslServer::from_files<C: AsRef<Path, K: AsRef<Path>>(cert: C, key: K)`

src/iron.rs
```diff
@@ -67,32 +67,40 @@ impl Default for Timeouts {
...
-    #[cfg(feature = "ssl")]
-    Https {
-        /// Path to SSL certificate file
-        certificate: PathBuf,
-        /// Path to SSL private key file
-        key: PathBuf
-    }
+    Https,
+    #[doc(hidden)]
+    __Extensible,
```
untitaker: Huh, why?

seanmonstar:
I waited on that issue for a few days to allow discussion, it seems it's wrapped up. I'm away from the computer today, but I can probably finish the release tomorrow.

src/iron.rs
seanmonstar: To allow adding new variants in the future.

I'd actually recommend making the enum part private, and expose function constructors instead.

```rust
pub struct Protocol(Proto);
enum Proto { .. }

impl Protocol {
    pub fn http() -> Protocol {
        Protocol(Proto::Http)
    }
}
```

examples/https.rs
sfackler: Updated! I haven't published the openssl and hyper-openssl releases yet, but assuming this looks good I will.

src/iron.rs
sfackler: Done

examples/https.rs
sfackler: Hmm, now that I think about it, I wonder if this example would be better off using native-tls instead?

sfackler:
Ok, I've switched the https example over to native-tls, which should make things nicer for OSX and especially Windows users.

untitaker:
Can we re-add the tests for this in Travis?

sfackler:
Done

untitaker:
Thanks!

PR #511, opened 2017-01-09
Add version to Request

sfackler:
Also add a private field to make it extensible in the future

### DIFF
--- a/src/request/mod.rs
+++ b/src/request/mod.rs
@@ -16,6 +16,9 @@ use method::Method;
 pub use hyper::server::request::Request as HttpRequest;
 use hyper::buffer;
 
+#[cfg(test)]
+use std::net::ToSocketAddrs;
+
 pub use self::url::Url;
 
 use {Protocol, Plugin, Headers, Set, headers};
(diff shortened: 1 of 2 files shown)


sfackler:
It might actually be nicer to make Request opaque and provide accessors, since a bunch of this stuff is stored directly in the hyper request.

untitaker:
It would be, but that's backwards-incompatible.

src/request/mod.rs
```diff
@@ -116,9 +124,26 @@ impl<'a, 'b> Request<'a, 'b> {
...
             method: method,
-            extensions: TypeMap::new()
+            extensions: TypeMap::new(),
+            version: version,
+            _p: (),
         })
     }
+
+    #[cfg(test)]
+    pub fn stub() -> Request<'a, 'b> {
```
untitaker: Why move this? IMO it's nice that code for testing is separated.

sfackler:
This is already a backwards-incompatible change since it's adding a field to a struct with no private fields.

#510 is a breaking change anyway, so I'm not sure why that matters?

src/request/mod.rs
sfackler: The _p field prevents a Request from being instantiated outside this module.

untitaker:
I guess this is backwards-incompatible if you pattern-match the entire struct, but I don't think that's a common thing to do, so I would've passed it off as okay.

Fair point about #510.

untitaker:
@sfackler feel free to write the API change to accessors in a new PR.

sfackler:
:+1: will do

PR #513, opened 2017-01-11
Remove ResponseBody

sfackler:
This originally existed to avoid having to wrap the Hyper Respnose in a
trait object, but since we're doing that now anyway we may as well avoid
the extra indirection

r? @untitaker


untitaker:
Thanks!

untitaker:
oh, cc @reem I think. I don't see a reason either why ResponseBody should persist, but perhaps there's a need for some reason we're missing.

PR #515, opened 2017-01-12
Remove error crate

sfackler:
This is a holdover from before the standard library Error could be
downcast.

Does this seem right @reem?

This may or may not be a breaking change - not sure how others would feel.


Hoverbear:
Seems reasonable for me... Definitely might be breaking though.

untitaker:
Yeah that's a breaking change. We can merge it into master, but not any of the maintenance branches.

untitaker:
Thanks!

src/error.rs
```diff
@@ -1,10 +1,9 @@
-use std::error::Error as StdError;
 use std::fmt;
 
 use modifier::Modifier;
 use {Response};
 
-pub use err::Error;
+pub use std::error::Error;
```
sfackler: We may want to cut this reexport. I only had it there as a (failed) attempt to keep the change roughly backwards compatible.

untitaker: Ok, feel free to make a PR

PR #516, opened 2017-01-18
Cleanup some clippy lints.

tomprince:
- Markdown in docstrings.
- Remove some needless clones.
- Get rid of pattern matching references to references.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -10,7 +10,7 @@ pub use hyper::error::Result as HttpResult;
 
 /// The type of Errors inside and when using Iron.
 ///
-/// IronError informs its receivers of two things:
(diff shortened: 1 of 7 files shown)


src/error.rs
```diff
@@ -36,7 +36,7 @@ pub struct IronError {
 }
 
 impl IronError {
-    /// Create a new IronError from an error and a modifier.
+    /// Create a new 'IronError' from an error and a modifier.
```
sfackler: These should be backticks I think.

tomprince: Yes, fixed.

sfackler:
Thanks!

PR #521, opened 2017-02-07
Make PathBuf set the MIME type based on the path

mbrubeck:
This now shares code with the Modifier impl for Path, which already sets the MIME type.


untitaker:
Neat, thanks!

PR #522, opened 2017-02-10
(doc) Explain how to use Body as Read in its documentation

robinst:
It wasn't obvious to me how to do that and a note like this in the
documentation would have helped me.

### DIFF
--- a/src/request/mod.rs
+++ b/src/request/mod.rs
@@ -146,7 +146,16 @@ impl<'a, 'b> Request<'a, 'b> {
     }
 }
 
-/// The body of an Iron request,
+/// The body of an Iron request.
+///
+/// To pass the body to other APIs that expect a `io::Read`, use a `&mut` reference:
+///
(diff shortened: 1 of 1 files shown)


untitaker:
I'm unsure about this. I don't think the API docs are the right place for this.

robinst:
It's one of the places where I looked for it. We can also put it in more than one place, that's not a bad thing for documentation. Where would you want to put it instead?

untitaker:
To be honest I don't really understand wny Iron needs to document this. You can't pass `request.body` directly because that would destructure the request. You can't pass `&request.body` because you need to modify the body to be able to read it.

This is one of those things that naturally results out of how ownership works in Rust. I don't think Iron's API docs are the correct place to put this. I suppose we could have an example just for this though.

robinst:
Sorry for keeping this open. I'll look into adding another example.

Issue #501, opened 2016-11-15
Port to Hyper's tokio branch


untitaker:
Rough plan:

* Add a new trait `AsyncHandler` with a method `handle_async`. There is a `impl<H: Handler> AsyncHandler for H` that just delegates an invocation of `Handler::handle(self, ...)` to a threadpool. (where is the threadpool stored?). The idea is that users can still impl synchronous handlers using `Handler`, but drop down to `AsyncHandler` if they really care about performance.

  * Important: Implement `AsyncHandler` for functions, etc
* All middlewares become async. Synchronous middlewares are not supported. `AroundMiddleware` takes `Box<AsyncHandler>` and returns `Box<AsyncHandler>`
* Everything else in Iron goes async as well.
* The response object will have to change somehow, probably. Hopefully response modifiers can abstract the changes away.

Hoverbear:
This will be a big job and introduce significant breaking changes.

ernestas-poskus:
If you plan to make middlewares async it might brake expected behavior where authorization or alike middleware must go first.

untitaker:
Middlewares would still be applied in order.

Hoverbear:
Right. By async we mean the middlewares themselves should support being async, so they should not block, for example. They would still be done in order.

hayd:
Hyper's tokio branch is now in master. https://github.com/hyperium/hyper/commit/2d2d5574a698e74e5102d39b9a9ab750860d92d1

untitaker:
#523 updates to async Hyper, for now just using a threadpool

MJDSys:
I've started digging into the meat of this issue (Async Iron on Async Hyper, based off of #523).  However, I'm running some lifetime issues.

Basically, when my ```async_handle``` function returns, I can't have references to either the handler or the request hanging around, as the response isn't constructed until later.  For requests, I've tried stepping around this by moving in a ```Request``` instead of passing a ```&mut Request```, and producing a ```Request``` as part of the output.  I'm not super happy about this though, as it leaks the details of the ```Chain``` into ```async_handle```'s interface.  Also there is now an ugly tuple to return.  The trait is currently:
```rust
/// `AsyncHandler`s are responsible for handling all requests by returning a future that yields the actual response.
pub trait AsyncHandler: Send + 'static {
    /// Produce a `Future` that handles the given request that produces an appropriate response or error.
    fn async_handle(&self, Request) -> BoxIronFuture<(Request, Response)>;
}

[...]

Issue #446, opened 2016-04-06
Hyper MIO Branch Compatibility


bfrog:
It would be great to see how, if at all, the mio branch of hyper will change things with Iron.

bfrog:
One thing that I've found useful is the ability to have per listener thread state. In my case the Sender side of std::sync::mpsc::channel's, but I could see other per thread data being useful. So perhaps something akin to iron/persistent but with a way to set it up per worker thread. Maybe even simply allowing the user to create an iron server per thread as hyper does now would allow per thread state to be created for handlers.

bfrog:
I spent some time investigating this and it really comes down to the Body readable at the moment is problematic. To avoid pushing all the async stuff into Iron the request body needs to either be entirely in memory or entirely on disk. As an alternative perhaps a continuation can be returned as a IronResult specifying it needs the Body to begin with. In all scenarios Body is no longer the same as it is in Iron 0.3

The in memory/on disk scenario I could see playing out basically where if the request body is extraordinarily large, its written to disk before being given. This is troublesome though because, what size is the maximum allowed? What's the cutoff for in memory or on disk?

The continuation method seems saner. Providing a closure as an IronResult, lets call it a BodyReader takes either the entirety or chunks of the body when needed...

So basically something like...

``` rust
fn some_handler(req: IronRequest) -> IronResult { 
     let request_handler = SomeHandler::new(req);
    IronResult::ReadBody(move |chunk| { request_handler.handle_chunk(chunk) }) 
}
```

Not quite as nice as the std::io::Read for sure, but without something akin to co-routines...

reem:
So thinking about this more, the ideal situation is:

- we have an in-memory buffer for reading data from the socket
- we read into this buffer whenever a chunk is available until it is full
- user code can drain from this buffer asynchronously
- when the buffer is full we stop reading from the socket until the user code has drained some data, then we resume (the logic here is a bit tricky, but it should work)

This way we don't have to deal with an ever-expanding in-memory buffer or resorting to a disk buffer.

In terms of how we actually expose this to the user, something like a specialized `eventual` Stream would probably be the best way to go without going full coroutines, which could also work.

reem:
In terms of how I'm thinking to land something like this: I'd like to do one more release of iron that will be the final release based off of sync hyper and that exposes a sync interface, then start another release for the asynchronous version. We can then maintain the synchronous version for some time before deprecating it.

bfrog:
Seems like a sensible way of doing things to me

[...]

PR #523, opened 2017-02-19
[RFC] [WIP] Async Iron with Hyper tokio

MJDSys:
This PR is the start of me working on porting Iron over to async Hyper/tokio (issue #501).  This RFC start just moves everything to compile on the new branch.

I don't know if you want to accept this PR (after whatever cleanup is necessary!) as a base so someone (maybe me if my time stays available) can start working on a proper async story for Iron.  If you prefer this PR develops into the full async support, that's fine too I'll just leave it open as I continue developing.

### DIFF
--- a/src/iron.rs
+++ b/src/iron.rs
@@ -2,11 +2,14 @@
 //! `Iron` library.
 
 use std::net::{ToSocketAddrs, SocketAddr};
+use std::sync::Arc;
 use std::time::Duration;
 
-pub use hyper::server::Listening;
-use hyper::server::Server;
-use hyper::net::{Fresh, SslServer, HttpListener, HttpsListener, NetworkListener};
+use futures::{future, Future, BoxFuture};
+use futures_cpupool::CpuPool;
+
+use hyper::{Body, Error};
+use hyper::server::Http;
 
 use request::HttpRequest;
 use response::HttpResponse;
@@ -116,84 +119,82 @@ impl<H: Handler> Iron<H> {
     ///
     /// The thread returns a guard that will automatically join with the parent
     /// once it is dropped, blocking until this happens.
-    pub fn http<A>(self, addr: A) -> HttpResult<Listening>
+    pub fn http<A>(self, addr: A) -> HttpResult<()>
         where A: ToSocketAddrs
     {
-        HttpListener::new(addr).and_then(|l| self.listen(l, Protocol::http()))
-    }
-
-    /// Kick off the server process using the HTTPS protocol.
-    ///
-    /// Call this once to begin listening for requests on the server.
-    /// This consumes the Iron instance, but does the listening on
-    /// another task, so is not blocking.
-    ///
-    /// The thread returns a guard that will automatically join with the parent
-    /// once it is dropped, blocking until this happens.
-    pub fn https<A, S>(self, addr: A, ssl: S) -> HttpResult<Listening>
-        where A: ToSocketAddrs,
-              S: 'static + SslServer + Send + Clone
-    {
-        HttpsListener::new(addr, ssl).and_then(|l| self.listen(l, Protocol::http()))
-    }
-
-    /// Kick off a server process on an arbitrary `Listener`.
-    ///
-    /// Most use cases may call `http` and `https` methods instead of this.
-    pub fn listen<L>(self, mut listener: L, protocol: Protocol) -> HttpResult<Listening>
-        where L: 'static + NetworkListener + Send
-    {
-        let handler = RawHandler {
-            handler: self.handler,
-            addr: try!(listener.local_addr()),
-            protocol: protocol,
+        let addr = addr.to_socket_addrs().unwrap().next().unwrap();
+        let handler = RawService{
+            addr: addr,
+            handler: Arc::new(self.handler),
+            protocol: Protocol::http(),
+            pool: CpuPool::new(self.threads),
         };
+        Http::new().bind(&addr.clone(), handler).and_then(|server| server.run())
+    }
+}
 
-        let mut server = Server::new(listener);
-        server.keep_alive(self.timeouts.keep_alive);
-        server.set_read_timeout(self.timeouts.read);
-        server.set_write_timeout(self.timeouts.write);
-        server.handle_threads(handler, self.threads)
+struct RawService<H> {
+    handler: Arc<H>,
+    addr: SocketAddr,
+    protocol: Protocol,
+    pool: CpuPool,
+}
+
+impl<H: Handler> ::hyper::server::NewService for RawService<H> {
+    type Request = HttpRequest;
+    type Response = HttpResponse;
+    type Error = ::hyper::Error;
+    type Instance = RawHandler<H>;
+
+    fn new_service(&self) -> Result<Self::Instance, ::std::io::Error> {
+        Ok(RawHandler{
+            handler: self.handler.clone(),
+            addr: self.addr.clone(),
+            protocol: self.protocol.clone(),
+            pool: self.pool.clone(),
+        })
     }
 }
 
 struct RawHandler<H> {
-    handler: H,
+    handler: Arc<H>,
     addr: SocketAddr,
     protocol: Protocol,
+    pool: CpuPool,
 }
 
-impl<H: Handler> ::hyper::server::Handler for RawHandler<H> {
-    fn handle(&self, http_req: HttpRequest, mut http_res: HttpResponse<Fresh>) {
-        // Set some defaults in case request handler panics.
-        // This should not be necessary anymore once stdlib's catch_panic becomes stable.
-        *http_res.status_mut() = status::InternalServerError;
+impl<H: Handler> ::hyper::server::Service for RawHandler<H> {
+    type Request = HttpRequest;
+    type Response = HttpResponse;
+    type Error = Error;
+    type Future = BoxFuture<Self::Response,Self::Error>;
 
-        // Create `Request` wrapper.
-        match Request::from_http(http_req, self.addr, &self.protocol) {
-            Ok(mut req) => {
-                // Dispatch the request, write the response back to http_res
-                self.handler.handle(&mut req).unwrap_or_else(|e| {
-                    error!("Error handling:\n{:?}\nError was: {:?}", req, e.error);
-                    e.response
-                }).write_back(http_res)
-            },
-            Err(e) => {
-                error!("Error creating request:\n    {}", e);
-                bad_request(http_res)
-            }
-        }
+    fn call(&self, req: Self::Request) -> Self::Future {
+        let addr = self.addr.clone();
+        let proto = self.protocol.clone();
+        let handler = self.handler.clone();
+        self.pool.spawn_fn(move || {
+            let mut http_res = HttpResponse::<Body>::new().with_status(status::InternalServerError);
+
+            match Request::from_http(req, addr, &proto) {
+                Ok(mut req) => {
+                    // Dispatch the request, write the response back to http_res
+                    handler.handle(&mut req).unwrap_or_else(|e| {
+                        error!("Error handling:\n{:?}\nError was: {:?}", req, e.error);
+                            e.response
+                    }).write_back(&mut http_res)
+                },
+                Err(e) => {
+                    error!("Error creating request:\n    {}", e);
+                    bad_request(&mut http_res)
+                }
+            };
+            future::ok(http_res)
+        }).boxed()
     }
 }
 
-fn bad_request(mut http_res: HttpResponse<Fresh>) {
-    *http_res.status_mut() = status::BadRequest;
-
-    // Consume and flush the response.
-    // We would like this to work, but can't do anything if it doesn't.
-    if let Ok(res) = http_res.start()
-    {
-        let _ = res.end();
-    }
+fn bad_request(mut http_res: &mut HttpResponse<Body>) {
+    http_res.set_status(status::BadRequest);
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -69,6 +69,8 @@ extern crate plugin;
 extern crate url as url_ext;
 extern crate num_cpus;
 extern crate conduit_mime_types as mime_types;
+extern crate futures;
+extern crate futures_cpupool;
 #[macro_use]
 extern crate lazy_static;
 
@@ -151,8 +153,8 @@ pub mod status {
 
 /// HTTP Methods
 pub mod method {
-    pub use hyper::method::Method;
-    pub use hyper::method::Method::*;
+    pub use hyper::Method;
+    pub use hyper::Method::*;
 }
 
 // Publicized to show the documentation
--- a/src/modifiers.rs
+++ b/src/modifiers.rs
@@ -12,10 +12,11 @@
 //! return status code:
 //!
 //! ```
+//! # use std::convert::Into;
 //! # use iron::prelude::*;
 //! # use iron::status;
 //! let r = Response::with(status::NotFound);
-//! assert_eq!(r.status.unwrap().to_u16(), 404);
+//! assert_eq!(404 as u16, r.status.unwrap().into());
 //! ```
 //!
 //! You can also pass in a tuple of modifiers, they will all be applied. Here's
@@ -160,17 +161,17 @@ impl Modifier<Response> for status::Status {
 
 /// A modifier for changing headers on requests and responses.
 #[derive(Clone)]
-pub struct Header<H: headers::Header + headers::HeaderFormat>(pub H);
+pub struct Header<H: headers::Header>(pub H);
 
 impl<H> Modifier<Response> for Header<H>
-where H: headers::Header + headers::HeaderFormat {
+where H: headers::Header {
     fn modify(self, res: &mut Response) {
         res.headers.set(self.0);
     }
 }
 
-impl<'a, 'b, H> Modifier<Request<'a, 'b>> for Header<H>
-where H: headers::Header + headers::HeaderFormat {
+impl<H> Modifier<Request> for Header<H>
+where H: headers::Header {
     fn modify(self, res: &mut Request) {
         res.headers.set(self.0);
     }
--- a/src/request/mod.rs
+++ b/src/request/mod.rs
@@ -1,20 +1,19 @@
 //! Iron's HTTP Request representation and associated methods.
 
-use std::io::{self, Read};
-use std::net::SocketAddr;
 use std::fmt::{self, Debug};
+use std::io::Cursor;
+use std::net::SocketAddr;
 
-use hyper::uri::RequestUri::{AbsoluteUri, AbsolutePath};
-use hyper::net::NetworkStream;
-use hyper::http::h1::HttpReader;
-use hyper::version::HttpVersion;
+use futures::Stream;
+
+use hyper::HttpVersion;
 
 use typemap::TypeMap;
 use plugin::Extensible;
 use method::Method;
 
-pub use hyper::server::request::Request as HttpRequest;
-use hyper::buffer;
+pub use hyper::Body;
+pub use hyper::server::Request as HttpRequest;
 
 #[cfg(test)]
 use std::net::ToSocketAddrs;
@@ -29,12 +28,12 @@ mod url;
 ///
 /// Stores all the properties of the client's request plus
 /// an `TypeMap` for data communication between middleware.
-pub struct Request<'a, 'b: 'a> {
+pub struct Request {
     /// The requested URL.
     pub url: Url,
 
     /// The originating address of the request.
-    pub remote_addr: SocketAddr,
+    pub remote_addr: Option<SocketAddr>,
 
     /// The local address of the request.
     pub local_addr: SocketAddr,
@@ -43,7 +42,7 @@ pub struct Request<'a, 'b: 'a> {
     pub headers: Headers,
 
     /// The request body as a reader.
-    pub body: Body<'a, 'b>,
+    pub body: Cursor<Vec<u8>>,
 
     /// The request method.
     pub method: Method,
@@ -57,7 +56,7 @@ pub struct Request<'a, 'b: 'a> {
     _p: (),
 }
 
-impl<'a, 'b> Debug for Request<'a, 'b> {
+impl Debug for Request {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         try!(writeln!(f, "Request {{"));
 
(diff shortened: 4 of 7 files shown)


untitaker:
This is excellent! I think for now we can definetly use a threadpool and maybe introduce async handlers at a later point as per #446.

untitaker:
However, the tests fail because the https parts fail.

MJDSys:
@untitaker It looks like hyper removed https support from the tokio branch.  If I disable the tls feature for now, is that ok?  I think any other option will take a lot more work + time.  I'll work on that along with the async support.

src/request/mod.rs
```diff
@@ -43,7 +42,7 @@ pub struct Request<'a, 'b: 'a> {
     pub headers: Headers,
 
     /// The request body as a reader.
-    pub body: Body<'a, 'b>,
+    pub body: Cursor<Vec<u8>>,
```
sfackler: This seems a bit scary - we'll want to be able to handle streaming request bodies.

MJDSys: Definitely!  I'm thinking that can just store a future for the body count, either using the hyper Chunk, or streaming out the bytes or some such directly.

That was just the easiest way to make it work/compile now.  I can see about making it nicer for this PR if you wish, or handle it later.  This was just very much geared to be a base to start on :)

untitaker: @MJDSys it is up to you if you want to continue this PR or leave it at that. No doubt it's a very useful start either way!

sfackler: I the hyper client's response interface might be a good thing to look at - I think it's basically a channel of byte buffers?

MJDSys:
I pushed a change that makes the body in the Request struct just hold the Stream from Hyper for now.  If Iron wants to wrap that in the future, that can always change.

There are two issues remaining before I can get an async Iron working:
  - [ ] Hyper removed SSL support, so I'm not sure how we want to handle it.  I think if another library exposes an ssl layer to tokio, it would be easy to wrap Hyper in that.  I'll investigate that later.
  - [ ] My first attempt at an AsyncHandler trait looked like: ```fn async_handle(&self, &mut Request) -> BoxIronFuture<Response>;```.  However, when trying to implement AsyncHandler for Handler, the compiler was not happy about the TypeMap in Response as Response needs to be Send (which TypeMap isn't).  Trying to change the TypeMap to a SendMap then breaks the integration with the plugin crate, and trying to change plugin to be generic over all the different TypeMaps is turning out to be very difficult.  I think a few projects rely on this functionality, so I'll continue to try chasing this down.

untitaker:
I think we should postpone the `AsyncHandler` to a different PR

MJDSys:
I've pushed some commits to bring back https support in Iron.  Due to the way dependency works, I'm not sure having the native-tls-example feature is, as native-tls now a direct dependency through tokio-tls.

I'm also not sure if the complexity of Iron::https is worth the simplification of Iron::listen.  My initial implementation of listen instead took a stream of futures that returned emitted the Io object.  listen could easily wait on the future without blocking the thread, but it meant the http function need to wrap every TcpStream in a future.  I've posted an initial version here: https://github.com/MJDSys/iron/blob/async2/src/iron.rs#L123 .  If you prefer that version, let me know and I'll update the last commit.

untitaker:
Before Hyper 0.10, the entire `https` function wasn't available without a SSL feature flag. I suppose it's time to bring this back and make `tokio_tls` an optional dependency.

The rest of your changes feels quite a bit over my head, perhaps somebody else can review them.

src/iron.rs
```diff
@@ -111,89 +121,138 @@ impl<H: Handler> Iron<H> {
...
     {
-        HttpListener::new(addr).and_then(|l| self.listen(l, Protocol::http()))
+        let addr = addr.to_socket_addrs()?.next().unwrap();
+
+        let core = Core::new().unwrap();
+        let handle = core.handle();
+
+        let sock = TcpListener::bind(&addr, &handle).unwrap().incoming();
+
+        return self.listen(sock, addr, Protocol::https(), core, handle);
```
untitaker: you probably meant `Protocol::http`here

MJDSys:
I fixed the typo, and readded the feature gate for ssl.  Let me know if there are any other changes wanted to land this PR.

In the mean time, I'll start thinking about async_handler again.

untitaker:
@reem @Hoverbear please review

MJDSys:
Ping!  I've pushed a new commit to make the branch work against the latest mater branch of Hyper.  Any other work needed before this can be merged?

I'm working on the proper async handler, but it's blocked currently by issues with the plugin crate, which I'm working on.

untitaker:
Again I'd like somebody else to review this as well...

MJDSys:
@untitaker Yep np. @reem @Hoverbear could you please review this PR?  Any comments/suggestions?

src/request/mod.rs
```diff
@@ -71,73 +69,72 @@ impl<'a, 'b> Debug for Request<'a, 'b> {
...
             version: version,
             _p: (),
         })
     }
 
+    /// Get the contents of the body as a Vec<u8>
+    ///
+    /// This consumes the body future and turns it into Vec<u8>.  Note this should not be called
+    /// from the main hyper thread, as it will potentially deadlock.
+    pub fn get_body_contents(&mut self) -> Option<Vec<u8>> {
```
reem: The best way to read the body would be similar to how it's done in extensions today, by reading into a `Vec`, storing the `Vec` in extensions, then returning a reference to it. You can still provide a separate method to "take" the `Vec` from the extensions, but you end up with friendlier `&self -> &[u8]` and `&mut self -> &mut [u8]` signatures.

MJDSys: 100% I wasn't happy with this api.  I was just trying to expose the body contents somehow, and had figured the details would change with the async api (especially since the async side wouldn't grab all the bytes before exposing the body).

For now, how about a function that reads the body into a Vec, stores it into extensions, and returns that reference.  Afterwards the function can continue returning that reference, so it isn't taken.  Does that sound good enough for now?  And Is get_body_contents a good enough name?

reem: That sounds great. I don't love the name, but I can't really think of anything better right now. `get_body` is easy to confuse with `.body`, and anything else is as long. Does anyone else have ideas? I'll keep thinking about it.

There is one further issue with this API: it encourages reading an unlimited amount of data over the network, which is basically never what you want to do. The body-parser crate gets around this by having a [default maximum length](https://github.com/iron/body-parser/blob/master/src/lib.rs#L63) and [accepting further configuration](https://github.com/iron/body-parser/blob/master/src/lib.rs#L49) via `extensions`. We could do that here instead, but the API is getting a little complex to go in core.

MJDSys: I've pushed an update that reads the body into an extension.  I didn't expose the key, as I didn't figure it was worth it.

Assuming iron is going to move to an async handler, I'm not sure of the value of this function anyway.  Calling it from the context of a future on the reactor's thread could deadlock the server.  I made it originally to provide a similar API to the original Body class in Iron, and we could just delete it frankly.

reem: I considered just deleting it but it's such a common thing to ask for in examples and such... I think we should just go with this implementation for now - after all we are still pre 1.0 and can remove it if it proves to be a source of problems.

MJDSys:
Ping!  I've updated this PR to work against the latest hyper changes.  Is there anything else blocking this merge?

sfackler:
A Hyper release? :P

MJDSys:
@sfackler Fair enough!  If that is the case, I'll just keep this PR working against latest master, but won't bother pinging until there is a release.  Let me know either way.

src/iron.rs
```diff
@@ -111,89 +128,139 @@ impl<H: Handler> Iron<H> {
...
     /// Call this once to begin listening for requests on the server.
-    /// This consumes the Iron instance, but does the listening on
-    /// another task, so is not blocking.
-    ///
-    /// The thread returns a guard that will automatically join with the parent
-    /// once it is dropped, blocking until this happens.
-    pub fn https<A, S>(self, addr: A, ssl: S) -> HttpResult<Listening>
-        where A: ToSocketAddrs,
-              S: 'static + SslServer + Send + Clone
+    #[cfg(feature = "ssl")]
```
sfackler: It seems very unfortunate to lock HTTPS support to native-tls.

src/iron.rs
```diff
@@ -111,89 +128,139 @@ impl<H: Handler> Iron<H> {
...
+                future::ok((sock, remote_addr))
+            }).then(|r| {
+                tx.send(r).map_err(|e| IoError::new(ErrorKind::Other, e))
+            }).and_then(|_| future::ok(()))
+        }).then(|_| future::ok(()));
+        handle.spawn(ssl_acceptor);
+
+        let listener = rx.then(|r| match r {
+            Ok(real_r) => real_r,
+            Err(e) => panic!(e),
```
sfackler: panic?

src/iron.rs
MJDSys: This shouldn't lock us into native-tls, just tokio-tls.  This still uses the tokio-tls trait TlsAcceptorExt, which should be implementable by any TLS crate (see: https://tokio-rs.github.io/tokio-tls/tokio_tls/trait.TlsAcceptorExt.html ).

If the tokio-tls dependency is unwanted, I could make our own version of that trait.  I think that implementation can even make using something implementing TlsAcceptorExt work.

Unfortunately, it does drag in a native-tls dependency to require tokio-tls.

sfackler: How can anything else implement those traits? The methods on them return concrete types (`ConnectAsync` and `AcceptAsync`) which do not have public constructors AFAIKT.

src/iron.rs
MJDSys: I believe this operation can "never" fail, so I included a panic to signify this.  But I can't find any documentation specifically saying this, just that the error type of r is the unit type.  I'll dig into this more sometime this week to verify.

src/iron.rs
MJDSys: Indeed!  I saw that there was a trait and naively assumed it would be generic over all possible tls implementations.  Ugh.

I agree limiting us to native-tls is unfortunate.  I'll change this to an iron specific trait for now, and implement it for tokio-tls.  Does that sound reasonable?

sfackler: Sure, though it seems like the trait should live in hyper-land so we don't need to redo all of the work in each Hyper consumer - @seanmonstar what's the state of TLS stuff in async Hyper?

seanmonstar: For Servers, the idea is that `hyper::server::Http` is a `ServerProto`, and can be wrapped in others. So, it doesn't care where the bytes are coming through, you can pipe them in from anything. An example of doing this using tokio-tls is:

```rust
let http = Http::new();
let tls = tokio_tls::proto::Server::new(http, native_tls_acceptor);
let tcp = tokio_proto::TcpServer::new(tls, addr);
// ...
```

src/iron.rs
bbigras: maybe use [std::unreachable](https://doc.rust-lang.org/std/macro.unreachable.html)

src/iron.rs
sfackler: Cool, thanks, that seems reasonable.

MJDSys: @sfackler I'm really not sure how to handle this now.  It doesn't appear there is a generic way to get a ServerProto to start a server.

If we don't mind being forever bound to using TCP (something my current implementation isn't reliant on), then I could make a trait with a single method that takes a ```hyper::server::Http``` and returns something implementing ```ServerProto```, which I can then use with TcpServer.  Then https can provide a default version that uses tokio-tls and http can have a simple identity function.

Alternately, we could get tokio-proto to make a trait to start a generic server.  I'm not sure how useful/plausible that is.  We could make our own for now while that is investigated.

Thoughts?

sunng87:
Hyper 0.11 has been released. What's the status of this?

untitaker:
Yeah we should get this merged and released soon. However I think this is outside of my expertise.

On 17 June 2017 11:17:44 GMT+02:00, Ning Sun <notifications@github.com> wrote:
>Hyper 0.11 has been released. What's the status of this?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/iron/iron/pull/523#issuecomment-309203894

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

MJDSys:
@sunng87 @untitaker I can update this PR to the official release probably sometime this weekend.  The only large remaining issue is how to handle SSL.  The current implementation ties us to openssl, but it would be nice to avoid that direct dependency.

I can take a stab at it again when I do the update, though if anyone has any suggestions I'd be happy to hear them!  Alternately, this could go in with the openssl dependency so people can start experimenting, and we fix it before release.  I'd be happy to help with that effort too.

MJDSys:
Ok, this PR is updated against the latest hyper release.  The only remaining problem is ssl.  My problem is that there is no generic tls mechanism for tokio, and tokio doesn't provide a generic way do deal with listeners.

Iron::listen at least allows for someone to construct their own tls handler on top of a TcpListener without having to deal with Hyper.  One improvement may be to return the server future, to avoid taking in the tokio reactor, but I'm not sure if that is useful.  I'm happy to implement any ideas people come up with.

One last idea: I could have Iron convert into an IronService (in replacement for RawService) (or have Iron implement NewService somehow), so someone could setup their own instance of hyper, and use Iron on top.  I'm not sure about this though.

seanmonstar:
The current tokio-proto design is built with some sort of idea of middleware, coming from 2 different sides of the server problem.

1. `ServerProto`: These are the things that receive some IO object, and process it into a `Request` and `Response` (not necessarily HTTP), and give those things to a `Service`. This is what `hyper::server::Http` is. A system of `ServerProto` middleware can exist. In fact, hyper's is one of them. There can exist other `ServerProto`s that are generic over one they wrap, such as the `TlsServer` in tokio-tls.

2. `Service`: These are the middlewares that process the `Request`/`Response` pair, not really dealing with the underlying IO object. Since Iron is built around the idea of middleware, it may make sense to be able to use any `Service` wrapper. This allows something like the following:
    
    ```rust
    let router = iron::Router::new(map_of_routes); // impl Service
    let timed = foo::Timed::new(thirty_seconds, router); // impl Service
    let log = iron::Log::new(log_opts, timed); // impl Service
    return log;
    ````

----

All this to say, you could make `Iron` implement `ServerProto`. Maybe it takes an `Http`, maybe that's not exposed, I don't really know. But that'd allow others to chose to use `TcpServer`, or `TlsServer`, or `UnixServer`, or `QuicServer`, or whatever.

If you wanted an "easy mode", you could make `Iron::listen` wrap itself in a `TcpServer` or something.

sunng87:
+1 for @seanmonstar . Iron as a web framework should be better to focus on framework part, leaving server implementation plugable for others. This is what we have in Python's WSGI ecosystem. There are a few frameworks (flask, django, bottle) and a few serves (fapws, mainheld).

seanmonstar:
I just realized that it may make sense to not even bother with `ServerProto`. Iron could be a collection of `Service`s. In which case, combining with hyper, it'd look like `Http::new().bind(addr, Iron::new(middlewares))`. And still with an `Iron::listen` easy mode if you prefer.

nicoburns:
> I just realized that it may make sense to not even bother with ServerProto. Iron could be a collection of Services. In which case, combining with hyper, it'd look like Http::new().bind(addr, Iron::new(middlewares)). And still with an Iron::listen easy mode if you prefer.

This would mirror how ExpressJS works: https://expressjs.com/en/4x/api.html#app.listen

Thomspoon:
Wondering what the progress on this is? I'd love to help where I can.

RobertWHurst:
So is this good to go?

RobertWHurst:
Is there any update here? My company is building a payments processor using Iron and we will need async for performance reasons.

MJDSys:
Hi all,

Sorry this took a while to come back to.  I was working to understand how tokio-proto worked with hyper (it all makes sense now, but there are some unclear parts from just reading documentation/implemented traits).  Anyways, now the Iron struct directly implements the NewService trait, which makes it trivial to work with hyper/tokio-proto, which can be seen in the http/https methods and the hello_custom_config example.

The around example no longer works as expected though, as the server is started as soon as http/https is called, and thus only a single server is started.  If it's ok that the http/https methods start the server and never return, I'll fix the example to configure LoggerMode based on an argument.

If wanted I can clean up the git history some to reduce the number of commits, but I figured getting the newest version of the code out was more important.

Also, what is the status on #533?  Should that land before this?  As I mentioned, that shouldn't be a problem and I can easily fix up my history to be formatted appropriately after it lands.

dnp1:
Hey guys, 
I would like this feautre, How can I help on this?

dnp1:
I'm trying to use the @MJDSys branch in my project.
So, I am migrating the core bundle libraris (params, bodyparser) to fit async

iori-yja:
I love to look forward this change merged, however, it will break vast of existing codes, so I suggest to change its version.

Also, I found a bug (will describe soon after) and I guess there'll be several regression; so this change should be merged into a new branch for being tested and reviewed by more people.

iori-yja:
This change has a bug on `BodyReader` implementation. Here is a reproducable code.

## symptom

```
fn stream(_: &mut iron::Request) -> iron::IronResult<iron::Response> {
    let bufreader = BufReader::new(fs::File::open("/dev/urandom").unwrap());
    Ok(iron::Response::with((status::Ok, response::BodyReader(bufreader))))
}
```
(Full code is [here](https://github.com/iori-yja/iron-streaming-example/blob/use-async-iron/src/main.rs) )

This code is expected to return random streaming response as the iron/iron version can. [a version](https://github.com/iori-yja/iron-streaming-example/blob/master/src/main.rs) using vanilla iron can do this.

## cause

This change stops to use HttpResponse<Body>::start() due to moving hyper 0.11. https://github.com/iron/iron/pull/523/files#diff-6cfc163228cc31bed4a7214fa0b69300L142 and now consumes into buffer before return the first drop.

Here's a backtrace of the server above which is being returning a response for first request.
```
(gdb) bt
#0  0x0000000801ce0616 in memcpy () from /lib/libc.so.7
#1  0x00000000010c3e0b in <[T] as core::slice::SliceExt>::copy_from_slice ()
#2  0x00000000010d031d in alloc::slice::<impl [T]>::copy_from_slice ()
#3  0x00000000010ddc90 in <alloc::vec::Vec<T> as alloc::vec::SpecExtend<&'a T, core::slice::Iter<'a, T>>>::spec_extend ()
#4  0x00000000010ddde5 in <alloc::vec::Vec<T>>::extend_from_slice ()
#5  0x00000000010de533 in std::io::impls::<impl std::io::Write for alloc::vec::Vec<u8>>::write_all ()
#6  0x00000000010a712e in std::io::util::copy (reader=0x802621320, writer=&mut Write = {...}) at /checkout/src/libstd/io/util.rs:66
#7  0x000000000109f1e2 in <iron::response::BodyReader<R> as iron::response::WriteBody>::write_body (self=0x802621320,
    res=&mut Write = {...}) at /home/iorivur/.cargo/git/checkouts/iron-8bf7332c1b743992/abf4cb2/src/response.rs:65
#8  0x00000000010c2459 in iron::response::write_with_body (res=0x7fffdfffa0d8, body=Box<WriteBody> = {...})
    at /home/iorivur/.cargo/git/checkouts/iron-8bf7332c1b743992/abf4cb2/src/response.rs:143
#9  0x00000000010c1dec in iron::response::Response::write_back (self=Response = {...}, http_res=0x7fffdfffa0d8)
    at /home/iorivur/.cargo/git/checkouts/iron-8bf7332c1b743992/abf4cb2/src/response.rs:122
#10 0x000000000109e7c0 in <iron::iron::IronHandler<H> as tokio_service::Service>::call::{{closure}} ()
    at /home/iorivur/.cargo/git/checkouts/iron-8bf7332c1b743992/abf4cb2/src/iron.rs:236
```

## possible fix

hyper 0.11 is using `mpsc`'s sender and I think we can send back a stream by pair() in `hyper`. https://docs.rs/hyper/0.11.12/src/hyper/proto/body.rs.html#48

(I think it'll be better to merge this prior to decide to release so that everyone can improve this on iron/iron's board)

beatgammit:
Just another annoying ping. What's left to get this merged?

I'm building something with iron ATM and I'm getting close to wanting to release a proof of concept, so I'd like to make sure the async story is moving along before I commit to iron.

Is there anything I can do to help get this merged? I have some time in the next few weeks I think.

robinst:
@beatgammit I would recommend using another framework. At the moment, unfortunately iron is not actively maintained. I've raised a PR to add a note to the readme about that: #576

phlmn:
So, this has been sitting way too long. I just updated the PR to hyper 0.12.5.

I removed SSL for now, when tokio-tls is updated (https://github.com/tokio-rs/tokio-tls/issues/44) we can either add an example showing how to use it with the http API from hyper or build a convenience function.

src/lib.rs
```diff
@@ -141,17 +140,12 @@ pub mod url {
 }
 
 /// Status Codes
-pub mod status {
-    pub use hyper::status::StatusCode as Status;
-    pub use hyper::status::StatusCode::*;
-    pub use hyper::status::StatusClass;
-}
+pub use http::status as status;
+pub use http::status::StatusCode;
```
zgtm: Why not use

    pub use http::status::StatusCode as Status

and not import `http::status` at all? This would also spare us some renaming¹ in lots of other places.

As far as I can tell, `http::status` basically never used except for some occurences of `status::StatusCode`—which already is imported as `StatusCode` right now.

¹ Not that it really would spare us renaming, since the names have all been changed to ALL_CAPS, but still …

src/modifiers.rs
```diff
@@ -129,7 +131,7 @@ impl<'a> Modifier<Response> for &'a Path {
     /// Panics if there is no file at the passed-in Path.
     fn modify(self, res: &mut Response) {
         File::open(self)
-            .expect(&format!("No such file: {}", self.display()))
+            .unwrap_or_else(|_| panic!("No such file: {}", self.display()))
```
zgtm: Why is this better than `.expect(&format!("No such file: {}", self.display()))`? Is it just to spare the call to self.display?

src/request/mod.rs
```diff
@@ -1,49 +1,48 @@
 //! Iron's HTTP Request representation and associated methods.
 
-use std::io::{self, Read};
-use std::net::SocketAddr;
+use std::u16;
```
zgtm: What's this doing here? ^.^ I think this line can safely be dropped.

src/modifiers.rs
phlmn: Probably because you don't call `format` when there is no error

src/lib.rs
phlmn: Umm... Like hyper? :D Seems like a good idea

phlmn: Ok hyper just exports StatusCode, I think we should also do that.

src/modifiers.rs
RobertWHurst: @phlmn I also don't understand this change. `format!` can be used for all sorts of use cases, I'm not sure I follow why it can't be used to build a string for an `expect` call.

phlmn: @RobertWHurst I did not write that line and I don't think there is any noticeable difference, but: If you write `expect(&format(…))` the `format(…)` is always executed while `unwrap_or_else(|_| panic!(…))` only executes a format if the `Result` is an `Err`.

examples/get_set_headers.rs
```diff
@@ -6,9 +6,9 @@ struct DefaultContentType;
 impl AfterMiddleware for DefaultContentType {
     // This is run for every requests, AFTER all handlers have been executed
     fn after(&self, _req: &mut Request, mut resp: Response) -> IronResult<Response> {
-        if resp.headers.get::<iron::headers::ContentType>() == None {
+        if resp.headers.get(iron::headers::CONTENT_TYPE) == None {
             // Set a standard header
-            resp.headers.set(iron::headers::ContentType::plaintext());
+            resp.headers.insert(iron::headers::CONTENT_TYPE, iron::headers::HeaderValue::from_static(iron::mime::TEXT_PLAIN.as_ref()));
```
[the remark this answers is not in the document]
phlmn: But then there we don't check if it's a valid content type… What about:
```
resp.headers.insert(iron::headers::CONTENT_TYPE, iron::mime::TEXT_PLAIN.as_ref().parse().unwrap());
```

examples/content_type.rs
```diff
@@ -1,46 +1,49 @@
...
-    let content_type = mime::Mime(iron::mime::TopLevel::Application, iron::mime::SubLevel::Json, vec![]);
-    Ok(Response::with((content_type, status::Ok, "{}")))
+    let content_type = "application/json".parse::<mime::Mime>().unwrap();
+    Ok(Response::with((content_type, StatusCode::OK, "{}")))
 }
 
 fn main() {
     let args: Vec<String> = env::args().collect();
-    let variant_index = if args.len() > 1 { args[1].parse().unwrap() } else { 1 };
+    let variant_index = if args.len() > 1 {
```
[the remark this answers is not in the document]
phlmn: I won't change this, since we should simply format the whole project with rustfmt.

src/modifiers.rs
```diff
@@ -63,7 +64,7 @@ use response::{WriteBody, BodyReader};
 impl Modifier<Response> for Mime {
     #[inline]
     fn modify(self, res: &mut Response) {
-        res.headers.set(headers::ContentType(self))
+        res.headers.insert(headers::CONTENT_TYPE, self.as_ref().parse().unwrap()); // TODO: error handling?
```
phlmn: What do we do with these errors?

zgtm: I think in this case, there is no error handling necessary, as a Mime should always parse into a valid string for HeaderValue.

src/modifiers.rs
```diff
@@ -91,7 +92,7 @@ impl Modifier<Response> for String {
 impl Modifier<Response> for Vec<u8> {
     #[inline]
     fn modify(self, res: &mut Response) {
-        res.headers.set(headers::ContentLength(self.len() as u64));
+        res.headers.insert(headers::CONTENT_LENGTH, (self.len() as u64).to_string().parse().unwrap()); // TODO: error handling?
```
zgtm: Here, `(self.len() as u64).into()` would be enough.

src/modifiers.rs
```diff
@@ -179,7 +180,7 @@ pub struct Redirect(pub Url);
 impl Modifier<Response> for Redirect {
     fn modify(self, res: &mut Response) {
         let Redirect(url) = self;
-        res.headers.set(headers::Location(url.to_string()));
+        res.headers.insert(headers::LOCATION, url.to_string().parse().unwrap()); // TODO: error handling?
```
zgtm: Here and 10 lines below: Here, also, the strings should always be valid HeaderValues. Thus, no error handling necessary.

phlmn: `parse()` could error here since only visible ASCII characters are permitted:  https://docs.rs/hyper/0.12.7/hyper/header/struct.HeaderValue.html#method.from_str

PR #527, opened 2017-03-30
Async handler

MJDSys:
Continuing with my work on #501, this adds an initial AsyncHandler trait to iron.  I'm not sure this is the right path, so I'm posting this PR to start discussions of this.

Current problems with this API:
 - [ ] AsyncHandler moves the Request, instead of taking a mutable reference.  I really can't see a way around this though while still implementing Chain.  The problem is keeping the reference to the Request alive for long enough, as async_handler returns immediately (and ending the borrow) while the middleware calls don't happen until after async_handler returns (ie when the futures complete).
 - [ ] AsyncHandler and AfterMiddleware both need to return the Request, since they are passed the request.  While not so bad for the middleware, for AsyncHandler it makes no sense (I created a response from the request, why do I need to return it?)  Maybe the response should contain the request instead?  I'm really unsure about this.
 - [ ] IronError now contains the Request as well, since the error path may need it.  This seems horrible to me, and makes dealing with errors in an AroundMiddleware a nightmare.  Alternately I could have the error path return the error alongside the request, but that doesn't seem nice either.
 - [ ] This PR avoids making everything Send by avoiding BoxFuture/.boxed, but I'm not sure it's worth the resulting ugliness.  However, requiring all the futures to be Send also means making some elements Sync (due to impl requirements on Arc regarding Send).  Though if we don't mind requiring Sync, the other changes are trivial AFAICT.
 - [ ] Also, I use boxed futures everywhere.  I think I could actually remove most of them by using an associative type on AsyncHandler + making the Chain modifiers consume the Chain + some templates on Chain.  However, I'm not sure how worthwhile that is now without impl Trait.  I'd like to take a shot at it sometime, so that the core Iron has as little overhead as possible.  Note I haven't benchmarked this, and it maybe better to keep this version.  I suspect the Boxes are a net loss though.
 - [ ] Due to the Request handling, the itry/iexpect macros now require the Request.  That's ... less then ideal.
 - [ ] Handler is now required to be wrapped in an Arc.  Not sure how much you care about that.  It's needed to send multiple references to the Handler to different threads in the cpu pool with unknown lifetimes.
 - [ ] The examples are still mostly Handlers, instead of AsyncHandler.  Not sure if you want any Handler examples, but I can change that once the rest of the API is locked down.
 - [ ] I have some concerns with my naming, but we can deal with that after.  I wasn't sure how much you cared about compatibility, considering this breaks everything.

In good news, the Chain tests are a little more reliable, as the tests now detect when a stage wasn't executed when it should have been (as opposed to when catch was run).  Regardless of what happens to this PR, I can pull that out if its wanted.

Comments/Suggestions/Concerns appreciated!


src/request/mod.rs
```diff
@@ -71,107 +68,89 @@ impl<'a, 'b> Debug for Request<'a, 'b> {
...
+    /// Get the contents of the body as a Vec<u8>
+    ///
+    /// This consumes the body future and turns it into Vec<u8>.  Note this should not be called
+    /// from the main hyper thread, as it will potentially deadlock.
+    pub fn get_body_contents(&mut self) -> &Vec<u8> {
+        if let Some(reader) = self.body.take() {
+            let body = reader.wait().fold(Vec::new(), |mut v, input| { v.extend_from_slice(&input.unwrap()); v });
+            self.extensions.insert::<RequestBodyKey>(body);
+        }
+        self.extensions.get::<RequestBodyKey>().unwrap()
```
alsuren: I'm only just starting to get the hang of async rust, but I suspect that this will eventually want to be refactored into an async version (which stores a shared future [1] in self.extensions and returns a clone of it), and a sync version that calls the async function and waits on its result.

[1] http://alexcrichton.com/futures-rs/futures/future/trait.Future.html#method.shared

PR #530, opened 2017-04-11
Adding a from_str method

dorfsmay:
This is to allow the use of the FromStr trait to create a Url directly from string, and for example use:

    let u = "https://en.wikipedia.org/wiki/Rust_(programming_language)".parse().unwrap();
    let redirect = iron::modifiers::Redirect(u);

instead of:

    let u = iron::Url::parse("https://en.wikipedia.org/wiki/Rust_(programming_language)").unwrap();
    let redirect = iron::modifiers::Redirect(u);

### DIFF
--- a/src/request/url.rs
+++ b/src/request/url.rs
@@ -1,6 +1,7 @@
 //! HTTP/HTTPS URL type for Iron.
 
 use url::{self, Host};
+use std::str::FromStr;
 use std::fmt;
 
 /// HTTP/HTTPS URL type for Iron.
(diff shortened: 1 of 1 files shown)


src/request/url.rs
```diff
@@ -137,6 +138,13 @@ impl AsMut<url::Url> for Url {
     fn as_mut(&mut self) -> &mut url::Url { &mut self.generic_url }
 }
 
+impl FromStr for Url {
+    type Err = String;
+    #[inline]
+    fn from_str(input: &str) -> Result<Url, Self::Err> {
+        Ok(Url::parse(input).unwrap())
```
Hoverbear: Please don't use `unwrap()` here.

dorfsmay: Ooops! Removed it. Thanks.

Hoverbear:
Looks good, I agree this is an improvement! Do you think we should add a test as well? :)

dorfsmay:
@Hoverbear When I looked at adding tests, I realised they'd all be exactly the same as for Url parse().

I guess we could add one simple positive + one simple negative test to ensure the code is not broken, then rely on parse() for the rest? Let me know what you think.

Hoverbear:
Yeah that's kind of what I thought too. :S I guess it's fine as is.

Hoverbear:
Thanks for this contribution! :) We really appreciate it!

PR #531, opened 2017-04-12
adding two simple tests for from_str

dorfsmay:
Before I saw your merge (thanks!) I thought one simple positive and one simple negative test would be good for when that particular piece of code is changed. No need to do anything more complicated given that there are already tests by parse() itself.


Hoverbear:
Excellent, thank you. This is exactly what I hoped for. :)

Issue #532, opened 2017-04-12
Why not use rustfmt?


dorfsmay:
Is there a reason why iron/iron code does not comply with rustfmt format?

It means that those of us who use rustfmt plug-ins have to disable the plug-ins from our editors before we can work on this project. It also open space for discussions about style...

I'm happy to create a Pull Request for this, but first want to know if there's a reason for not using it.

untitaker:
Feel free to open a PR and cc me

Hoverbear:
I'm concerned this will make merging #523 etc very hard.

dorfsmay:
@Hoverbear Would make sense to cancel this PR (there really was no worked involved), put a moratorium on big PRs until the current ones are being merged, and start this initiative again after that?

untitaker:
agreed with @dorfsmay, this could be done in either order. It doesn't matter tbh.

MJDSys:
@dorfsmay @untitaker I'm happy to do whatever as well.  If you wish to merge #533 first, I can rebase on top of that, maintaining the rustfmt runs.  In fact I suspect there is some git trickery I could do to make that trivial :).

Or we can merge #523, then redo #533 on top.  Whatever works for the iron project.

phlmn:
Since #523 is merged, I will now run rustfmt on all files :)

PR #533, opened 2017-04-13
Ran rustfmt on all .rs files

dorfsmay:
As per #532.
CC:  @untitaker 

All I did here is literally `cargo fmt`, `cargo test`, `git -u`.

Ideally, all new code should be formatted with rustfmt to avoid the constant reformatting between devs who use rustfmt and devs who don't. Git hooks maybe...

### DIFF
--- a/examples/404.rs
+++ b/examples/404.rs
@@ -4,8 +4,7 @@ use iron::prelude::*;
 use iron::status;
 
 fn main() {
-    Iron::new(|_: &mut Request| {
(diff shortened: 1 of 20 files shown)


robinst:
This should probably use the RFC style using `rustfmt-nightly` instead. But that's not ready yet (it's still changing quite a bit): https://github.com/rust-lang-nursery/rustfmt#rustfmt--

phlmn:
Since this is fairly outdated I made a new PR (#581).

PR #543, opened 2017-05-10
(doc) Add echo example

robinst:
Shows how to read request body and echo it back in a response, including
error handling.


Hoverbear:
Thanks for this! We need more examples. :)

robinst:
Agreed! I wonder if there's a nice way to make them more visible from the docs here: https://docs.rs/iron/0.5.1/iron/

(I didn't know about the examples before, and the docs itself are really minimal.)

Hoverbear:
Could just put a link to the examples directory in the frontpage of the docs maybe!

PR #544, opened 2017-05-11
Replace conduit-mime-types dep by mime_guess

robinst:
A couple of reasons:

* conduit-mime-types does not look actively maintained, its last commit was 2 years ago
* It depends on the deprecated rustc-serialize for parsing JSON at runtime
* mime_guess uses a compile-time hash map, which means we also don't need lazy_static anymore either

I split it into two commits: The first one adds tests for the existing functionality. The second one replaces the implementation while keeping the tests.

Here's the result of running [cargo tree](https://github.com/sfackler/cargo-tree) before and after:

```diff
@@ -1,6 +1,4 @@
 iron v0.5.1 (file:///Users/rstocker/Projects/rust/iron)
-├── conduit-mime-types v0.7.3
-│   └── rustc-serialize v0.3.24
 ├── hyper v0.10.10
 │   ├── base64 v0.5.2
 │   │   └── byteorder v1.0.0
@@ -23,8 +21,14 @@
 │       │   │   └── matches v0.1.4 (*)
 │       │   └── unicode-normalization v0.1.4
 │       └── matches v0.1.4 (*)
-├── lazy_static v0.2.8
 ├── log v0.3.7 (*)
+├── mime_guess v1.8.1
+│   ├── mime v0.2.3 (*)
+│   ├── phf v0.7.21
+│   │   └── phf_shared v0.7.21
+│   │       ├── siphasher v0.2.2
+│   │       └── unicase v1.4.0 (*)
+│   └── unicase v1.4.0 (*)
 ├── modifier v0.1.0
 ├── num_cpus v1.4.0 (*)
 ├── plugin v0.2.6
```

### DIFF
--- a/src/modifiers.rs
+++ b/src/modifiers.rs
@@ -135,8 +135,9 @@ impl<'a> Modifier<Response> for &'a Path {
             .expect(&format!("No such file: {}", self.display()))
             .modify(res);
 
-        let mime_str = MIME_TYPES.mime_for_path(self);
-        let _ = mime_str.parse().map(|mime: Mime| res.set_mut(mime));
+        if let Some(mime) = mime_for_path(self) {
+            res.set_mut(mime);
+        }
     }
 }
 
@@ -193,5 +194,28 @@ impl Modifier<Response> for RedirectRaw {
     fn modify(self, res: &mut Response) {
         let RedirectRaw(path) = self;
         res.headers.set(headers::Location(path));
+    }
+}
+
+fn mime_for_path(path: &Path) -> Option<Mime> {
+    let mime_str = MIME_TYPES.mime_for_path(path);
+    mime_str.parse().ok()
+}
+
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_mime_for_path() {
+        assert_eq!(mime_for_path(Path::new("foo.txt")),
+                   Some("text/plain".parse().unwrap()));
+        assert_eq!(mime_for_path(Path::new("foo.jpg")),
+                   Some("image/jpeg".parse().unwrap()));
+        assert_eq!(mime_for_path(Path::new("foo.zip")),
+                   Some("application/zip".parse().unwrap()));
+        assert_eq!(mime_for_path(Path::new("foo")),
+                   Some("text/plain".parse().unwrap()));
     }
 }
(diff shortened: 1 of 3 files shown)


src/modifiers.rs
```diff
@@ -195,3 +192,26 @@ impl Modifier<Response> for RedirectRaw {
         res.headers.set(headers::Location(path));
     }
 }
+
+fn mime_for_path(path: &Path) -> Mime {
+    guess_mime_type_opt(path)
+        .unwrap_or_else(|| Mime(TopLevel::Text, SubLevel::Plain, vec![]))
```
robinst: I kept the fallback the same as before: https://github.com/conduit-rust/mime-types/blob/master/src/lib.rs#L48

mime_guess would fall back to `application/octet-stream` instead, which might be the better default. But not sure if you wanna change that in Iron.

robinst:
@SkylerLipthay It looks like you introduced this feature, mind having a look? Would be appreciated!

SkylerLipthay:
Looks great, thank you for the detailed pull request. I think we'll leave the default MIME type as `text/plain` just because this would otherwise be a breaking change and given my current capacity in this project I'm not comfortable making that call. Also I'm not sure how other frameworks handle it... Anyway, this can of course be changed in the future if deemed necessary. Thanks again!

robinst:
Thanks!

robinst:
Oh, I was looking for when the last release was and noticed that there's a `0.5-maintenance` branch. Should I have made the PR against that branch for it to be included in the next 0.5 release?

SkylerLipthay:
Hmm. Well, that branch is quite behind `master` and it seems that all new commits over the past 4 months have gone straight into `master`. So I think you're good.

The question of "when is the next version scheduled for release?" is one I'm curious about myself... I don't know what this process looks like these days...

robinst:
Ok, thanks. Maybe @untitaker or @Hoverbear: It would be cool to get a 0.6.0 release with the changes in master since 0.5.1: https://github.com/iron/iron/compare/0.5.1...master :tada:

(Seems like iron using async hyper is still some time away, so it's probably worth having some more releases before that.)

Hoverbear:
@skylerlipthay I don't think we really schedule them. I'd be OK with having a new version published though.

robinst:
I opened a new issue about getting a release out.

Issue #528, opened 2017-04-07
The content-length of the Response does not follow RFC 7230


futursolo:
Currently, Iron always sends a content length header with the client when there's no body.

However, as per [RFC 7230 [section 3.3.2]](https://tools.ietf.org/html/rfc7230#section-3.3.2), there're certain cases that the web server should not send a content length header or the content length in the header does not match the actual length of the content.

    A server MAY send a Content-Length header field in a 304 (Not
    Modified) response to a conditional GET request (Section 4.1 of
    [RFC7232]); a server MUST NOT send Content-Length in such a response
    unless its field-value equals the decimal number of octets that would
    have been sent in the payload body of a 200 (OK) response to the same
    request.

For such a case, Iron always responds `0`, even override manually.

Iron Version: 0.5.1
Rust Version: Stable 1.16.0

phlmn:
Should be fixed by #523.

PR #560, opened 2017-09-22
HTTP 204 should have no body, not even Content-Length

kw217:
RFC7230 s3.3.2 says:

"A server MUST NOT send a Content-Length header field in any response with a status code of 1xx (Informational) or 204 (No Content)."

It also says that 304 is only allowed Content-Length in certain circumstances.

Currently Iron adds a `Content-Length: 0` header whenever there is no body. This is incorrect for 1xx and 204 responses and many 304 responses. This fix changes that behaviour, to strip any `Content-Length` header for 1xx and 204 responses, allow it but not add it for 304, and otherwise add it.

Edit: I discovered this because a Scala unmarshaller I'm using is refusing to parse 204 responses from Iron for this reason.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -9,7 +9,7 @@ use plugin::Extensible;
 use modifier::{Set, Modifier};
 use hyper::header::Headers;
 
-use status::{self, Status};
+use status::{self, Status, StatusClass};
 use {Plugin, headers};
 
 pub use hyper::server::response::Response as HttpResponse;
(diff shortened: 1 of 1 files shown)


robinst:
I think this would fix #528, right?

kw217:
Yes, this fixes #528 .

clarcharr:
Shouldn't this be included in Hyper, not Iron?

kw217:
Arguably it should go in Hyper _as well_, but the code in Iron should do the right thing too. I think this is fixed in Hyper 0.11 anyway, but Iron is still using Hyper 0.10.

phlmn:
Should be fixed with #523

kw217:
Thanks, that's fine.

PR #561, opened 2017-10-03
(doc) fix ssh key generating command error

INORIA:
Hello!

When I wanted to generate a ssh key, I got the following error.

```
Invalid password argument "mypass"
Error getting passwords
```

`pass:` may fix this issue.


untitaker:
Thanks!

Issue #526, opened 2017-03-10
Https listener is constructed with http protocol


frjonsen:
As seen here: https://github.com/iron/iron/blob/master/src/iron.rs#L137

When constructing a https server, Protocol::http() is used. This makes things like the iron-routers url_for! not work properly, because it constructs a http:// url even when using https.

phlmn:
Thanks for fixing that! :)

PR #565, opened 2017-11-08
patch for issues #526.

segfo:
Fixed an issue where the url_for method of iron-router does not work when using HTTPS listener.

The relevant problem is as follows.
https://github.com/iron/iron/issues/526

best regards


segfo:
I mistook a branch. Close

Issue #551, opened 2017-06-23
Release 0.6.0 with changes on master


robinst:
(From the comments here: https://github.com/iron/iron/pull/544#issuecomment-303621983)

It would be cool to get a 0.6.0 release with the changes in master since 0.5.1 (from January): https://github.com/iron/iron/compare/0.5.1...master

The following would be worth releasing:

* Dependency cleanups: https://github.com/iron/iron/compare/0.5.1...master#diff-80398c5faae3c069e4e6aa2ed11b28c0
* API change to use `std::error::Error`: https://github.com/iron/iron/compare/0.5.1...master#diff-9a269209bed069ffb36d5565a91c9840
* Documentation improvements

(It would be before hyper 0.11, as changing Iron to be based on async hyper will probably take a while.)

**Update:** iron 0.6.0 has been released. Now the iron dependency needs to be bumped to "0.6" in all the other iron crates (and released):

- [x] [router](https://github.com/iron/router): https://github.com/iron/router/pull/141
- [x] [mount](https://github.com/iron/mount): https://github.com/iron/mount/pull/95
- [x] [persistent](https://github.com/iron/persistent): https://github.com/iron/persistent/pull/64 / https://github.com/iron/persistent/pull/65
- [x] [bodyparser](https://github.com/iron/body-parser) (depends on persistent): https://github.com/iron/body-parser/pull/79
- [x] [urlencoded](https://github.com/iron/urlencoded) (depends on bodyparser): https://github.com/iron/urlencoded/pull/72
- [x] [params](https://github.com/iron/params) (depends on bodyparser and urlencoded) https://github.com/iron/params/pull/39
- [x] [iron-test](https://github.com/iron/iron-test) (depends on router, urlencoded) https://github.com/iron/iron-test/pull/47
- [x] [staticfile](https://github.com/iron/staticfile) (depends on mount, iron-test) https://github.com/iron/staticfile/pull/100
- [x] [logger](https://github.com/iron/logger)

While we're at it, we should bump other dependencies as well if needed.

robinst:
@reem @untitaker ?

robinst:
See here for why this is being held up (problems with permissions on crates.io): https://github.com/iron/urlencoded/issues/68

Hoverbear:
Ping @reem

cetra3:
What's the status of this request? Looks like the urlencoded stuff has been resolved?

untitaker:
[...]

PR #576, opened 2018-02-08
README: Add note about maintenance status of iron

robinst:
The last release of iron (0.6.0) took months to do and required a lot of pushing
from an outsider (me), see #551. Given that, I think it's fair to say that iron
is unmaintained and at this point I would advise new people not to use it.

If any of that changes, we can remove the notice again.

CC @untitaker

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+# NOTE: Iron is not actively maintained at the moment, please consider using a different framework
+
 Iron
 ====
 


sunng87:
There is a badge for the maintenance status, https://doc.rust-lang.org/cargo/reference/manifest.html

> # Maintenance: `status` is required Available options are `actively-developed`,
# `passively-maintained`, `as-is`, `none`, `experimental`, `looking-for-maintainer`
# and `deprecated`.
maintenance = { status = "..." } 

Of course an unmaintained repo won't get chance to put this on crates.io 😝

By the way, I [forked iron](https://github.com/sunng87/zhelezo) and picked up #523 from @MJDSys , fixed some deprecation warnings.  Not sure if I could manage to get it to right direction, but I definitely love the architecture and have some ideas to try.

remram44:
Probably a lot of people are happy/interested to take over, it would be really sad to see such a prominent framework die because the team doesn't have release access or time. Maybe get some more people on board?

Otherwise forking might be an option.

untitaker:
Thanks. I personally would recommend forking iron. Access to resources/services is scattered and not well-organized.

robinst:
For anyone considering forking:

* I think one of the things that's annoying about releasing iron at the moment is that it's scattered across so many repos, and each one has to be released separately. I would look at consolidating them into one repo with a cargo workspace.
* Probably keep the version the same across all the crates
* If you're gonna have new crate names, probably give them all a common prefix (instead of just `router`, etc)

sunng87:
A few ideas I want to try in my fork:

* **Iron-API**: Seek to create a relatively stable API for Request, Response and Middleware. App developers and middleware developers will use this API and never care about actual server impl. It can be great if the API is independent to underlying implementation, for example, does not involve hyper types like Body, while keep the performance. Consider it as the serde crate for Iron.
* **Iron-Server**: Keep track with latest Hyper/Tokio/Futures development to ensure backend solid and up to date.
* **Iron-middlewares**: keep some essential middlewares maintained and released together just like the stdlib.

Issue #579, opened 2018-07-06
Move core packages to a single repository


phlmn:
The release of version 0.6.0 took very long. As mentioned in https://github.com/iron/iron/pull/576#issuecomment-364627847 the fact that there are **11!!** repositories that needed to be updated is one reason.

I think it would be nice to move all 'core' repositories into the main repository and use cargo workspaces to keep them separated as crates.

The affected repositories are:

- [x] iron
- [x] router
- [x] logger
- [ ] iron-sessionstorage
- [ ] staticfile
- [ ] iron-test
- [ ] params
- [ ] urlencoded
- [ ] body-parser
- [x] mount
- [x] persistent

I would like to hear other opinions on that.

sunng87:
Since iron has an extensible architecture, even if we move all these crates into a workspace, those third-party crates like handlebars-iron still need to manually update when iron version changed. 

One idea comes to me is to create an relatively stable iron-api crate to let both core and middlewares to depend on. So we can update iron internals without affecting the ecosystem. Or perhaps we can have different implementation of iron, the hyper one and actix one. It's like serde for serde_json.

phlmn:
@sunng87 While I totally agree that the api should become stable, I don't think we will get there too soon. The next step probably is to update to a new hyper version with futures, which will break the api again…

I'm not sure if it is necessary to move the api to a different crate. Would you like to open an issue proposing that? There we should collect what would need to be in an api crate and see if it would be reasonable.

Nevertheless I also would put the api crate in this repository as a workspace. 😃 

Being able to use iron with different http servers would be awesome (further discussion should be in your issue #261), this might be possible when we have moved the api to the future based approach.

sunng87:
Yes I agree. By updating hyper to 0.12, we can take advantage of the new `http` crate, and we can probably  consider using it as our Request/Response type.

By iron-api I think it consists with:

* Request/Response
* Middleware traits

We can really think about this after hyper 0.12 adoption.

zgtm:
@sunng87: I'm not sure if that would actually buy you anything. Most of the repositories depending on iron use the types that iron re-exports from hyper (for status codes, methods, headers, …). Those will probably change in the future with upcoming hyper versions – and all dependent crates still need to be updated.

PR #591, opened 2018-10-19
Add mount, persistent and router as workspaces

phlmn:
First steps for #579


zgtm:
I think the examples should be moved to the `iron` package in the workspace in order to have them runnable via `cargo run --example …`. (Similar to the examples in the other packages.)

zgtm:
`mount` package is missing in `Cargo.toml`

Issue #594, opened 2018-12-13
'cargo test' fails / rustc 1.31.0 (stable)


sergey-melnychuk:
```
$ cargo test

error: use of deprecated item 'http::Uri::port': use `port_part` instead                                                                                                                                                                 
  --> src/request/mod.rs:95:28                                                                                                                                                                                                           
   |                                                                                                                                                                                                                                     
95 |                 (host, uri.port())                                                                                                                                                                                                  
   |                            ^^^^                                                                                                                                                                                                     
   |                                                                                                                                                                                                                                     
note: lint level defined here                                                                                                                                                                                                            
  --> src/lib.rs:6:24                                                                                                                                                                                                                    
   |                                                                                                                                                                                                                                     
6  | #![cfg_attr(test, deny(warnings))]                                                                                                                                                                                                  
   |                        ^^^^^^^^                                                                                                                                                                                                     
   = note: #[deny(deprecated)] implied by #[deny(warnings)]                                                                                                                                                                              
                                                                                                                                                                                                                                         
error: aborting due to previous error                                                                                                                                                                                                    
                                                                                                                                                                                                                                         
error: Could not compile `iron`.
```

sergey-melnychuk:
Friendly ping! :smile:

@phlmn 
@zgtm

PR #595, opened 2018-12-13
(fix) Use port_part() insead of deprecated port()

sergey-melnychuk:
https://github.com/iron/iron/issues/594


phlmn:
Thanks! 😊
And sorry, that it has taken so long... 😥

PR #599, opened 2019-01-22
Use source instead of cause in Error trait

dvic:
This should fix the nightly build (which is currently failing due to the `cause` deprecation warning).


phlmn:
Thanks! 🙂

PR #604, opened 2019-02-15
Fix query strings

zgtm:
This PR fixes iron completely ignoring query strings.

I guess the regression was introduced with some changes in the hyper `uri` module that no longer seems to include the query string in the `path` section of the request URL. I did not have time to look into the exact cause yet, though.


phlmn:
Was the query part of the path before?

renannprado:
@zgtm @phlmn is this patch correct? can it be merged?

phlmn:
Yeah, this can be merged :)

PR #607, opened 2019-03-18
fix simple_routing.rs example

ahxxm:
StatusCode was moved to mod status, and reuses hyper::status::StatusCode


ahxxm:
i see... iron on crates.io was outdated

PR #611, opened 2019-06-30
(doc) Fix usage of status on response timer.

alekhrycaiko:
Hi,

I was interested in getting going with Iron and gave your example code a try. I noticed that when I tried to use the Response timer example, I ran into an error with the way status code was imported and then consumed. I've made a change to the README that fixes this issue. Hopefully this will help someone in the future :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ extern crate iron;
 extern crate time;
 
 use iron::prelude::*;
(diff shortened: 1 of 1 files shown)


phlmn:
Hi,
looks like you tried to use the examples from the master branch with the released version 0.6. You can find examples for version 0.6 here: https://github.com/iron/iron/tree/0.6-maintenance.

PR #614, opened 2019-08-13
fix nightly build of router crate

miller-time:
closes https://github.com/iron/router/issues/148

- update route-recognizer (implicitly)
- add `dyn` to trait objects

### DIFF
--- a/router/src/router.rs
+++ b/router/src/router.rs
@@ -14,9 +14,9 @@ use recognizer::{Match, Params};
 
 pub struct RouterInner {
     // The routers, specialized by method.
-    pub routers: HashMap<method::Method, Recognizer<Box<Handler>>>,
+    pub routers: HashMap<method::Method, Recognizer<Box<dyn Handler>>>,
     // Routes that accept any method.
-    pub wildcard: Recognizer<Box<Handler>>,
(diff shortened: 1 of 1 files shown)


router/Cargo.toml
```diff
@@ -10,6 +9,6 @@ documentation = "http://ironframework.io/doc/router/index.html"
 keywords = ["iron", "web", "http", "routing", "router"]
 
 [dependencies]
-route-recognizer = "0.1"
+route-recognizer = "0.1.13"
```
phlmn: There should be no need for specifying a patch version, `cargo update` should be enough.

miller-time:
@phlmn I think "router" needs to be added to:

https://github.com/iron/iron/blob/cf056ea5e8052c1feea6141e40ab0306715a2c33/Cargo.toml#L3

should I add that to this PR?

phlmn:
@miller-time Yep, noticed that, I've just pushed the fix (https://github.com/iron/iron/commit/73fbcec1a5d57d9af0899147047a36ffd7a5cc16).

miller-time:
rebased :nerd_face:

phlmn:
beautiful ❤️

PR #620, opened 2019-12-06
Put query string into request URL

### LINKED ISSUE #617: iron::params not parsing query string parameters in 0.6.1 UrlEncodedQuery seems to work with same example.
I'm having trouble getting the example in the documentation to work. Even the examples throw errors. UrlEncodedQuery parses fine, no problems. Is Params still supported?

Url: 
http://localhost:5000/myApiEndpoint?name=jason&age=2

Handler:
fn getStatesByCountryId(req: &mut Request) -> IronResult<Response>{ ...

Code:

        println!("Params = {:?}", req.get_ref::<Params>());

        match req.get_ref::<UrlEncodedQuery>() {
            Ok(ref hashmap) => println!("Parsed GET request query string:\n {:?}", hashmap),
            Err(ref e) => println!("{:?}", e)
        };

Output:

Params = Err(InvalidPath)
Parsed GET request query string:
 {"name": ["jason"], "age": ["2"]}

renannprado:
For some reason query strings were no longer being taken into consideration
and the routes would get no query string value at all if they wanted to.
So this fixes #617 and adds a test to ensure it won't come back.

Several compilation errors related to try! macro were fixed as well.

### DIFF
--- a/iron/src/request/mod.rs
+++ b/iron/src/request/mod.rs
@@ -56,13 +56,13 @@ pub struct Request {
 
 impl Debug for Request {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        try!(writeln!(f, "Request {{"));
+        writeln!(f, "Request {{")?;
 
-        try!(writeln!(f, "    url: {:?}", self.url));
-        try!(writeln!(f, "    method: {:?}", self.method));
-        try!(writeln!(f, "    local_addr: {:?}", self.local_addr));
+        writeln!(f, "    url: {:?}", self.url)?;
+        writeln!(f, "    method: {:?}", self.method)?;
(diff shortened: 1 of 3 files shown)


renannprado:
@phlmn could you please review?

renannprado:
@phlmn any news here? is there anyone else I could ask for a review?

renannprado:
Nobody seems to have interest in this, closing.

phlmn:
Sorry, there is not much going on here... But the change looks good, so I'm happy to merge it.

robinhundt:
This code seems to be broken. On the current master the added test case fails with:
```
thread 'request::test::test_request_with_query_string' panicked at 'assertion failed: `(left == right)`
  left: `Some("param=value?param=value")`,
 right: `Some("param=value")`', iron/src/request/mod.rs:238:9
```

PR #623, opened 2019-12-09
(doc) fix broken logger link

sezna:
Two links point to the old rust doc urls. This PR updates those to their newer docs.rs urls. Additionally, fixed some grammar here and there.


sezna:
what needs to get done after approval for merging?

phlmn:
Nothing :) the approver is just no maintainer

PR #624, opened 2019-12-11
Fix 'examples' and their warnings

berlincount:
Some typos and depreacation warnings addressed.


berlincount:
hmm, let's ping @phlmn ;)

PR #645, opened 2025-03-17
docs: repaired a damaged badge

operagxoksana:


operagxoksana:
I couldn't find a solution to fix the Travis badge. Is it possible to remove it altogether?

